PR #2, opened 2022-12-28
MRD improved analysis and automatic notebook

kenissur:


ugvc/mrd/mrd_utils.py
```diff
@@ -80,6 +80,17 @@ def _get_hmer_length(ref: str, left_motif: str, right_motif: str):
     return x[np.argmax(x[:, 1].cumsum() >= left_motif_len + 1), 1]
 
 
+def _safe_tabix_index(vcf_file: str):
+    try:
+        pysam.tabix_index(vcf_file, preset="vcf", force=True)  # make sure vcf is indexed
+        if (not os.path.isfile(vcf_file)) and os.path.isfile(vcf_file + ".gz"):  # was compressed by tabix
+            vcf_file = vcf_file + ".gz"
```
doron-st, line 87: I don't follow this.
you try to index the vcf file, but if it doesn't exist, and it's .gz exists, why don't you try to index the .gz?
Anyway returning anything from an index function sounds like a confusing side-effect.
Can't we just trust the input to be .gz?

ugvc/mrd/mrd_utils.py
```diff
@@ -437,13 +451,26 @@ def intersect_featuremap_with_signature(
...
         in case the file already exists and function executed with no overwrite=True
+    ValueError
+        If input output_intersection_file does not end with .vcf or .vcf.gz
 
     """
+    # parse the file name
+    if output_intersection_file is None:
+        featuremap_name = _get_sample_name_from_file_name(featuremap_file, split_position=0)
+        signature_name = _get_sample_name_from_file_name(signature_file, split_position=0)
+        type_name = "" if is_matched is None else ".matched" if is_matched else ".control"
```
doron-st: OMG. for me this is hard to read.
I would prefer using match,case https://www.freecodecamp.org/news/python-switch-statement-switch-case-example/ if we're using python 3.10

If not,  even a long if elsif else would be better for me

doron-st: Also is_matched seems to function as non-boolean argument, so perhaps it shold be an enum instead?

doron-st (approved):
Just some bonus comments

ugvc/mrd/mrd_utils.py
kenissur: If the input is a .vcf file this function compresses it to .vcf.gz, then indexes it. What would make this clearer to read?

ugvc/mrd/mrd_utils.py
kenissur: I will change to a longer if-elif-else

I think it's just an optional boolean input. If you have an optional int input that is allowed to be None, do you call it something other than int?

doron-st: ok, not critical, your call.

PR #7, opened 2023-01-12
[BIOIN-732] Correct genotyping using imputation

gilhornung:
There are some pylint issues I need to settle, but I am pushing this so we can move forward


test/system/test_correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,59 @@
...
+from test import get_resource_dir, test_dir
+
+import pandas as pd
+
+from ugvc.pipelines import correct_genotypes_by_imputation
+
+
+class TestCorrectGenotypesByImputation:
+    inputs_dir = get_resource_dir(__file__)
+    test_inputs_dir = f"{test_dir}/resources/system/test_correct_genotypes_by_imputation"
```
doron-st: doesn't get_resource_dir gibes this to you?

test/system/test_correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def test_correct_genotypes_by_imputation(self, tmpdir):
+        os.makedirs(tmpdir, exist_ok=True)
+        with open(os.path.join(tmpdir, "chrom_to_cohort_vcf.json"), "w") as f:
```
doron-st, line 22: do you need the path to be absolute in this json?
Otherwise I think it is simpler to save the json as test input.

test/system/test_correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,59 @@
...
+                "--temp_dir",
+                str(tmpdir),
+                "--epsilon",
+                "0.01",
+            ]
+        )
+        # Check that the numbers of variants which were changed are the same
+        df0 = pd.read_csv(os.path.join(self.test_inputs_dir, "out_counts.csv"), index_col=0)
+        df1 = pd.read_csv(os.path.join(tmpdir, "output_counts.csv"), index_col=0)
+        assert df1.equals(df0)
```
doron-st, line 60: nice test!

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+        required=True,
+        type=str,
+    )
+    ap_var.add_argument(
+        "--stats_file",
+        help="Name of csv file to output counts of variants processed",
+        required=False,
+        type=str,
+    )
+    ap_var.add_argument(
```
doron-st: Is using tmpdir a good practice for modules to be integrated into docker/cromwell?
I usually save the outputs in an output_dir.

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+        i = 0
+        for a2 in range(num_alt + 1):
+            for a1 in range(a2 + 1):
+                gr_ar[i, 0] = a1
+                gr_ar[i, 1] = a2
+                i += 1
+    return gr_ar
+
+
+def phred(p) -> np.ndarray:
```
doron-st: type of p would be helpful

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+    q = -10 * np.log10(np.array(p, dtype=np.float))
+    return q
+
+
+def unphred(q) -> np.ndarray:
```
doron-st: ditto, for all following functions

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+    gt_list.sort(key=lambda e: (e is None, e))
+    return tuple(gt_list)
+
+
+def different_gt(gt1, gt2) -> bool:
+    out = sort_gt(gt1) != sort_gt(gt2)
+    return out
+
+
+def _f_along_allele(allele_i, gt_ar, f_hom, f_het):
```
doron-st: document please

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+        help="Verbosity: ERROR, WARNING, INFO, DEBUG",
+        required=False,
+        default="INFO",
+    )
+    ap_var.add_argument("--add_counting_category", help="Is the input a result of mutect", action="store_true")
+
+    return ap_var
+
+
+def _command_to_subset_vcf(chrom: str, vcf: str, outfile: str):
```
doron-st: thumbs up for access definition.
Why not make it private though? 
two underscores

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+
+                out_vcf_obj.write(rec)
+
+        subprocess.check_output(f"bcftools index -t {out_vcf}", shell=True)
+
+        return dict(counters)
+
+
+def run(argv: list[str]):
+    """Correct a vcf based on imputation"""
```
doron-st: genotypes?

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+
+    # Load the mapping between the chromosome names and the reference cohort files
+    with open(args.chrom_to_plink, "r", encoding="utf-8") as json_file:
+        chrom_to_plink = json.load(json_file)
+
+    subset_files = {chrom: os.path.join(args.temp_dir, f"subset.{chrom}.vcf.gz") for chrom in chrom_to_cohort.keys()}
+    subset_vcf_cmds = [
+        _command_to_subset_vcf(chrom, args.input_vcf, subset_files[chrom]) for chrom in chrom_to_cohort.keys()
+    ]
+    sp.run_parallel(subset_vcf_cmds, args.threads_for_contig)
```
doron-st: nicely set.

consider rearanging like this:
files_1 = 
files_n = 
commands_1 = 
commands_n = 
run_parallel_1 
run_parallel_n

I find it easier to read, but perhaps a matter of taste

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+    def run_add_imputation_to_vcf(chrom):
+        out = add_imputation_to_vcf(
+            beagle_anno_vcf=beagle_anno_files[chrom],
+            out_vcf=os.path.join(args.temp_dir, f"add_imp.{chrom}.vcf.gz"),
+            epsilon=args.epsilon,
+            add_counting_category=args.add_counting_category,
+        )
+        return out
+
+    counts_per_chrom = Parallel(n_jobs=args.threads_for_contig, verbose=10)(
```
doron-st: multi-threaded python exists?!

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+
+def modify_stats_with_imp(
+    pl_tup, num_alt: int, ds_ar: np.ndarray, current_gt, epsilon: float
+) -> tuple[tuple[int], int, tuple[int]]:
+    """
+    Modifies the PL values bases on the results from the imputation (the ALT dose, DS)
+    Input:
+    pl_tup - A tuple with PLs (as in the vcf)
+    num_alt - number of ALT alleles
+    ds_ar - DS values in numpy array
```
doron-st: an example of ds_ar would help here

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+    current_gt - The genotype (GT) as in the vcf
+    epsilon - The weight of the imputation in determining the new PL. A number between 0 and 1.
+    Output:
+    The new PLs, The new GQ and the new genotype (GT)
+    """
+    gt_ar = genotype_ordering(num_alt)
+
+    # Modify the PLs with the imputation data
+    # f_het ane f_hom are the probabilities for het or hom, for each allele, capped by the epsilon
+    f_het = np.maximum(epsilon, np.minimum(2 - ds_ar, 1 - epsilon))
```
doron-st: This is difficult to understand.
If I may suggest:

def bound_arr(arr, min, max):
   return np.maximum(max, np.minimum(min, arr))
   
def convert_ds_arr_to_genotype_imputation_priors(ds_arr, epsilon):
""" 
convert beagle DS array to genotype imputation priors
epsilon=0.99, ds=1 -> f_het = 0.99, f_hom = 0.01
epsilon=0.99, ds=2 -> f_het = 0.01, f_hom = 0.00
"""

\# lowercound ds at 1, since we are only considering genotype correction (in the range [1,2])
ds_ar = np.minimum(ds_ar, 1)

f_het = 2 - ds_ar  # ds:2->f:0, ds:1->f:1
f_hom = ds_ar - 1 # ds:2 -> f:1, ds:1 -> f:0

f_het = bound_arr(f_het, 1 - epsilon, epsilon)
f_hom = f_het = bound_arr(f_hom, 1 - epsilon, epsilon)

return f_het, h_hom

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+    # value (het/hom) in the relevant position.
+    f_allele_ar = np.stack([_f_along_allele(i + 1, gt_ar, f_hom, f_het) for i in range(num_alt)],
+                           axis=1)
+    # The probablity for each genotype is the max across all alleles of that genotype. When there are
+    # missing values (i.e. where there is no imputation data on an allele, use the lowest cap).
+    f_gt_ar = np.amax(np.nan_to_num(f_allele_ar, nan=epsilon), axis=1)
+    # I don't change the PL for the hom-ref genotype, so the f is set to 1
+    f_gt_ar[0] = 1
+
+    pl_unphred = unphred(pl_tup)
```
doron-st: I think I lost you here...
I would need to talk about this in order to understand

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,477 @@
...
+
+        pl_format = in_vcf_obj.header.formats["PL"]
+        in_vcf_obj.header.formats.add(
+            id="PL0",
+            number=pl_format.number,
+            type=pl_format.type,
+            description=pl_format.description + " (DeepVariant output)",
+        )
+        if add_counting_category:
+            in_vcf_obj.header.info.add(id="CAT", type="String", number=1, description="Category, as counted")
```
doron-st: Is this a new info field?
what does it represent?

doron-st (commented):
Very nice work, some cosmetics comments, and one part where I lost track

test/system/test_correct_genotypes_by_imputation.py
gilhornung: I think I prefer the path to be absolute. Less prone to problems this way

ugvc/pipelines/correct_genotypes_by_imputation.py
gilhornung: I separated between the dir where all the intermediate vcfs are generated (and there are many of them) and the path to the final output vcf. I think it is a good practice. In essence, the arguments can point to the same output dir, and all outputs will be saved. The name "temp_dir" doesn't mean it has to be `/tmp/`.

ugvc/pipelines/correct_genotypes_by_imputation.py
gilhornung: Fixed

ugvc/pipelines/correct_genotypes_by_imputation.py
gilhornung: Yes. We use joblib's Parallel a lot

ugvc/pipelines/correct_genotypes_by_imputation.py
gilhornung: Added a better explanation

ugvc/pipelines/correct_genotypes_by_imputation.py
gilhornung: I agree this is the most confusing part. I documented the calculation in a confluence page, and I now reference to it. Furthermore, I followed you suggestion and tried to make it clearer by creating the _convert_ds_to_genotype_imputation_priors function (although it is different than your suggestion)

ugvc/pipelines/correct_genotypes_by_imputation.py
gilhornung: It is mainly for debugging purposes. I am counting variants that were (or not) affected by imputation, and here I add the category of the count directly to the vcf.

gilhornung: I changed the description. I hope it is clearer now

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,495 @@
...
+        i = 0
+        for a2 in range(num_alt + 1):
+            for a1 in range(a2 + 1):
+                gr_ar[i, 0] = a1
+                gr_ar[i, 1] = a2
+                i += 1
+    return gr_ar
+
+
+def phred(p: tuple[int] | np.ndarray) -> np.ndarray:
```
ilyasoifer: I think this should have been implemented in ugvc already and if not should belong to more general code

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,495 @@
...
+    q = -10 * np.log10(np.array(p, dtype=np.float))
+    return q
+
+
+def unphred(q: tuple[int] | np.ndarray) -> np.ndarray:
+    p = np.power(10, -np.array(q, dtype=np.float) / 10)
```
ilyasoifer: Can move to more general code

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,495 @@
...
+    --columns INFO/IMP,INFO/DR2,FORMAT/DS -o {outfile} -O z  {vcf} && \
+    bcftools index -t {outfile}
+    """
+    return cmd
+
+
+def genotype_ordering(num_alt: int) -> np.ndarray:
+    # Returns a numpy array with the order of the genotypes (based on the section Genotype Ordering in the VCF spec).
+    # Each row is a genotype, and the columns are the alleles.
+    gr_ar = np.full([int((num_alt + 2) * (num_alt + 1) / 2), 2], fill_value=-1, dtype=int)
```
ilyasoifer: Should go to `else`.

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,495 @@
...
+        help="Name of csv file to output counts of variants processed",
+        required=False,
+        type=str,
+    )
+    ap_var.add_argument(
+        "--temp_dir",
+        help="Path to temporary dir",
+        required=False,
+        type=str,
+        default="/tmp",
```
ilyasoifer: I think there is an environment variable that ensures that the temporary directory exists. A bit more stable, since this does not necessarily exist on cromwell

ilyasoifer (commented):
Looks good! 
Few questions inline and here: 
What is the function of the test_correct_genotypes../script? I did not see it being used. 
I think that all code that works on genotypes should also be split out into something more general. I see it being used in other places too

ugvc/pipelines/correct_genotypes_by_imputation.py
gilhornung: I moved it to vcfbed/vcftools.py

test/system/test_correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,87 @@
...
+import numpy as np
+
+from ugvc.pipelines import correct_genotypes_by_imputation as corgen
+from ugvc.vcfbed.vcftools import genotype_ordering
+
+class TestCorrectGenotypesByImputation:
+    inputs_dir = get_resource_dir(__file__)
+    resources_dir = get_resource_dir(__file__)
+
+    def test_convert_ds_to_genotype_imputation_priors(self):
```
doron-st: this is great.
It should be in test/unit/ though

doron-st: Also if you can separate each case to it's own test function that would be even better

test/system/test_correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,87 @@
...
+        # Homozygous biallelic variant
+        ds_ar = np.array([2])
+        gt_ar = genotype_ordering(1)
+        num_alt = 1
+        epsilon = 0.01
+        assert np.all(corgen._convert_ds_to_genotype_imputation_priors(ds_ar, gt_ar, num_alt, epsilon) == np.array([1,0.01,0.99]))
+        # Heterozygous biallelic variant
+        ds_ar = np.array([1])
+        gt_ar = genotype_ordering(1)
```
doron-st: this 1 is num_alt defined below.
It would be good to use the variable instead of a constant

test/system/test_correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,87 @@
...
+
+import pandas as pd
+import numpy as np
+
+from ugvc.pipelines import correct_genotypes_by_imputation as corgen
+from ugvc.vcfbed.vcftools import genotype_ordering
+
+class TestCorrectGenotypesByImputation:
+    inputs_dir = get_resource_dir(__file__)
+    resources_dir = get_resource_dir(__file__)
```
doron-st, line 18: one of these we probably don't need now?

ugvc/pipelines/correct_genotypes_by_imputation.py
```diff
@@ -0,0 +1,455 @@
...
+    return cmd
+
+def _f_along_allele(allele_i: int, gt_ar: np.ndarray, f_hom: np.ndarray, f_het: np.ndarray) -> np.ndarray:
+    ar = (np.any(gt_ar == allele_i, axis=1) & (gt_ar[:, 0] == gt_ar[:, 1])) * f_hom[allele_i - 1] + (
+        np.any(gt_ar == allele_i, axis=1) & (gt_ar[:, 0] != gt_ar[:, 1])
+    ) * f_het[allele_i - 1]
+    return ar
+
+
+def _convert_ds_to_genotype_imputation_priors(
```
doron-st, line 189: looks good!

doron-st (approved):
Almost there, 
mind my pendatic self, few tiny comments on the tests

test/system/test_correct_genotypes_by_imputation.py
gilhornung: Problem with test/unit is that they do not refer to the files in the `pipelines` directory. Only system refers to pipeline.

PR #9, opened 2023-01-15
Fixed github action

ilyasoifer:
base conda environment started to collide with environment.yml that was installed as an update
i changed the workflow so that we install environment.yml in genomics.py3 environment


noam-ultima (approved):
@ilyasoifer Not sure I really understand your change, but I see this is working as github action so approved.

PR #11, opened 2023-01-17
vcfeval_flavors

doron-st:
alternative ways to penalize allele and genotype errors in vcfeval
The parameter controlling the penalty is -p (--penalty)
-p 2: usual vcfeval
-p 1: penalize wrong-allele/genotype only once (remove 0.5 of such fps and fns)
-p 0: don't penalize wrong-allele/genotype (remove completely such fps and fns)
-p -1: reward a tp for calling wrong-allele/genotype

### DIFF
--- a/ugvc/pipelines/vcfeval_flavors.py
+++ b/ugvc/pipelines/vcfeval_flavors.py
@@ -48,15 +48,19 @@ def get_parser(argv: list[str]):
                              "-p 1: penalize wrong-allele/genotype only once (remove 0.5 of such fps and fns)\n" \
                              "-p 0: don't penalize wrong-allele/genotype (remove completely such fps and fns)\n" \
                              "-p -1: don't penalize, and also reward a tp for calling wrong-allele/genotype\n")
+    ap.add_argument('--var_type',
+                    type=str,
+                    choices=['snps', 'indels', 'both'],
+                    default='both',
+                    help='choose to analyze specific variant type (indels is much faster than snps)')
+
+
+
     return ap
     return args
 
 def count_vcf_lines(vcf_file):
-    v_iter = pysam.VariantFile(vcf_file)
-    count = 0
-    for variant in v_iter:
-        count += 1
-    return count
+    return int(subprocess.check_output(f'bcftools index -n {vcf_file}'.split()))
 
 def run(argv: list[str]):
     """Evaluate VCF against baseline, giving alternative penalty to wrong-alleles and genotype errors"""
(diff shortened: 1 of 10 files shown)


ugvc/pipelines/vcfeval_flavors.py
```diff
@@ -0,0 +1,113 @@
...
+    calls = args.calls
+    baseline = args.baseline
+    eval_reg = args.evaluation_regions
+    sdf = args.template
+    out_dir = args.output
+    penalty = args.allele_and_genotype_error_penalty
+    result = []
+    sp.print_and_run(f'rm -rf {out_dir}')
+    sp.print_and_run(f'rtg vcfeval -b {baseline} -c {calls} -e {eval_reg} -t {sdf} -o {out_dir}')
+    sp.print_and_run(f'bedtools intersect -a {calls} -b {eval_reg} -header > {out_dir}/input_in_hcr.vcf')
```
gilhornung: I would use bcftools view with the -R option. Probably the same, but it is best to stay in the bcftools framework. It handles weird edge cases better.

ugvc/pipelines/vcfeval_flavors.py
```diff
@@ -0,0 +1,113 @@
...
+    sp.print_and_run(f'bcftools view -f PASS {out_dir}/input_in_hcr.vcf > {out_dir}/input_in_hcr.pass.vcf')
+    sp.print_and_run(f'bedtools intersect -a {baseline} -b {eval_reg} -header > {out_dir}/gtr_in_hcr.vcf')
```
gilhornung: Same. bcftools view -R

ugvc/pipelines/vcfeval_flavors.py
```diff
@@ -0,0 +1,113 @@
...
+                        choices=[2, 1, 0, -1],
+                        default=0,
+                        help="-p 2: usual vcfeval, penalizes twice each wrong-allele/genotype errror\n" \
+                             "-p 1: penalize wrong-allele/genotype only once (remove 0.5 of such fps and fns)\n" \
+                             "-p 0: don't penalize wrong-allele/genotype (remove completely such fps and fns)\n" \
+                             "-p -1: don't penalize, and also reward a tp for calling wrong-allele/genotype\n")
+    return ap
+    return args
+
+def count_vcf_lines(vcf_file):
```
gilhornung, line 58: Consider using `bcftools index -n`, instead (pysam is sloooooooow)

ugvc/pipelines/vcfeval_flavors.py
```diff
@@ -0,0 +1,113 @@
...
+    sp.print_and_run(f'bedtools intersect -a {baseline} -b {eval_reg} -header > {out_dir}/gtr_in_hcr.vcf')
+    result.append('type tp fp fn precision recall f1')
+    for vt in ['indels', 'snps']:
+        for pref in ['input_in_hcr', 'gtr_in_hcr', 'input_in_hcr.pass']:
+            sp.print_and_run(f'bcftools view {out_dir}/{pref}.vcf --type {vt} > {out_dir}/{pref}.{vt}.vcf')
+        sp.print_and_run(f'bcftools view {baseline} --type {vt} > {out_dir}/gt.{vt}.vcf')
+        for pref in ['fp', 'tp', 'fn']:
+            sp.print_and_run(f'bcftools view {out_dir}/{pref}.vcf.gz --type {vt} > {out_dir}/{pref}.{vt}.vcf')
+        sp.print_and_run(f'bedtools subtract -A -a {out_dir}/fp.{vt}.vcf -b {out_dir}/gt.{vt}.vcf -header >  {out_dir}/fp.{vt}.clean.vcf;')
+        sp.print_and_run(f'bedtools subtract -A -a {out_dir}/fn.{vt}.vcf -b {out_dir}/input_in_hcr.pass.{vt}.vcf -header > {out_dir}/fn.{vt}.clean.vcf')
```
gilhornung: So you separate between "calling errors" and genotype errors, by looking at the same positions.
But what about multiallelic variants that appear in the same position? Or Cases where the ground truth variant has a different allele then what we have? (e.g. ground truth has AG insertion and we call AAG insertion, but they have the same position)

Also - consider using bcftools isec instead of bedtools

gilhornung (changes requested):
I'm mostly concerned about variants that appear in the same position, but with a different allele than the ground truth. e.g. multialleleic variants or onsertion of different lengths

ugvc/pipelines/vcfeval_flavors.py
doron-st: ok, will try it.

ugvc/pipelines/vcfeval_flavors.py
doron-st: I separate between calling errors (completely missing variant, or calling variant where there are none)
and other errors (genotype errors, or calling incorrect allele)

ugvc/pipelines/vcfeval_flavors.py
doron-st: great, I was looking for something like that!

gilhornung (approved):

PR #14, opened 2023-01-23
[BIOIN-816] Variant filtering now does not over-ride previous filtering, just adds to it

ilyasoifer:


ugvc/pipelines/filter_variants_pipeline.py
```diff
@@ -180,15 +180,13 @@ def run(argv: list[str]):
             )
             with pysam.VariantFile(args.output_file, mode="w", header=hdr) as outfile:
                 for i, rec in tqdm.tqdm(enumerate(infile)):
-                    pass_flag = False
                     if hmer_run[i]:
                         rec.info["HPOL_RUN"] = True
                     if args.model_file is not None:
                         if predictions[i] == "fp":
+                            if "PASS" in rec.filter.keys():
+                                del rec.filter["PASS"]
```
doron-st, line 188: can you explain the nature of these changes.
The context is not very clear to me.
What do you want to happenm and what happened before?

doron-st (approved):
Would appreciate a short explanation

TammushB (approved):

ilyasoifer:
@doron-st I just want to be able to apply filter_variants_pipeline to the VCF without overwriting other filter annotations.

PR #15, opened 2023-01-25
Gil [BIOIN-732] maintain order of original vcf

gilhornung:
Changed the way in which I concat the vcfs, such that the order of the contigs in the vcf's header will be maintained


test/unit/vcfbed/test_vcftools.py
```diff
@@ -189,3 +192,61 @@ def test_get_vcf_df_ignore_fields(self):
...
+                                                      header,
+                                                      output_vcf,
+                                                      tmpdir)
+        vcounts = []                                              
+        bcftools_out = subprocess.check_output(f"bcftools index -s {output_vcf}", shell=True)
+        bcftools_lines = bcftools_out.decode().strip().split("\n")
+        for line in bcftools_lines:
+            columns = line.split("\t")
+            vcounts.append(int(columns[2]))
+        print(vcounts)
```
ilyasoifer: Print?

test/unit/vcfbed/test_vcftools.py
```diff
@@ -189,3 +192,61 @@ def test_get_vcf_df_ignore_fields(self):
...
```
ilyasoifer: Print?

ugvc/vcfbed/vcftools.py
```diff
@@ -561,3 +563,47 @@ def genotype_ordering(num_alt: int) -> np.ndarray:
...
+    
+    Parameters
+    ----------
+    input_vcf: The main vcf which is being modified
+    new_data_json: A json file in the form {chromosome: vcf}, with the data that will be replaced in input_json
+    header_file: A file with a valid vcf header that will replace the current header. Make sure the new header fits
+                 the new data.
+    output_vcf: Name of the output vcf
+    tempdir: Name of temp dir
+    """
```
ilyasoifer, line 593: Can you add "Returns" please?

ugvc/vcfbed/vcftools.py
```diff
@@ -561,3 +563,47 @@ def genotype_ordering(num_alt: int) -> np.ndarray:
...
+    # Extract chromosome without imputation data, and reheader to fit the header of vcfs after imputation
+    untouched_contigs_vcf = os.path.join(tempdir, "untouched_contigs.vcf.gz")
+    subprocess.check_call(f"bcftools view -O z -o {untouched_contigs_vcf} -R {untouched_contigs_bed} {input_vcf}", shell=True)
+    untouched_contigs_rehead_vcf = os.path.join(tempdir, "untouched_contigs_rehead.vcf.gz")
+    subprocess.check_call(f"bcftools reheader -h {header_file} -o {untouched_contigs_rehead_vcf} {untouched_contigs_vcf}", shell=True)
+    # Merge
+    vcfs_to_merge = list(chrom2vcf.values())
+    vcfs_to_merge.append(untouched_contigs_rehead_vcf)
+    subprocess.check_call(f"picard MergeVcfs INPUT={' INPUT='.join(vcfs_to_merge)} OUTPUT={output_vcf}", shell=True)
+    subprocess.check_call(f"bcftools index -tf {output_vcf}", shell=True)
```
ilyasoifer, line 626: Are you sure that MergeVCFs does not generate index?

ugvc/vcfbed/vcftools.py
```diff
@@ -561,3 +563,47 @@ def genotype_ordering(num_alt: int) -> np.ndarray:
...
+            contigs_info[contig.name] = {"id": contig.id, "length": contig.length}
+    sorted_contigs = sorted(contigs_info.keys(), key=lambda k: contigs_info[k]["id"])
+    # bed file with the chromosomes that were not modified by imputation
+    untouched_contigs_bed = os.path.join(tempdir, "untouched_contigs.bed")
+    with open(untouched_contigs_bed, "w", encoding="utf-8") as regions_file:
+        sorted_contigs = sorted(contigs_info.keys(), key=lambda k: contigs_info[k]["id"])
+        for contig in sorted_contigs:
+            if contig not in chrom2vcf.keys():
+                regions_file.write(contig + "\t1\t" + str(contigs_info[contig]["length"]) + "\n")
+    # Extract chromosome without imputation data, and reheader to fit the header of vcfs after imputation
```
ilyasoifer, line 613: I think the issue (maybe) is that the imputation changes the header (it should not). Maybe this is the part that we should fix?

ilyasoifer (commented):
Hi, looks good, my main issue (like in the WDL part) is all the dealing with the header. Can you remind me why this is needed?

ugvc/vcfbed/vcftools.py
gilhornung: It does, but it is an old version of the index, and it does not support `bcftools index -s` which I use in the tests

ugvc/vcfbed/vcftools.py
gilhornung: I add more fields to the vcf, such as the PLs and GTs before the imputation (PL0, GT0 and GQ0) and the DS values that Beagle outputs. So there is a change of the header.

ugvc/vcfbed/vcftools.py
gilhornung: > Hi, looks good, my main issue (like in the WDL part) is all the dealing with the header. Can you remind me why this is needed?

When I modify the vcfs based on imputation I add new fields: The Beagle DS, the previous values of the PL, GT and GQ.

test/unit/vcfbed/test_vcftools.py
gilhornung: removed

ugvc/vcfbed/vcftools.py
gilhornung: Added

test/unit/vcfbed/test_vcftools.py
gilhornung: Fixed

ugvc/vcfbed/vcftools.py
ilyasoifer: But you use here `bcftools index -s`. I don't understand the claim

gilhornung: `bcftools index -s` uses an *existing* index to output statistics about the number of variants in each contigs. The test that I am running (unit test) uses `bcftools index -s`. The index that is produced by GATK does not support this option, so I have to recreate it using bcftools.
The time to generate the new index is negligible, so I don't see any added cost to this command.

ilyasoifer: OK, at least maybe add a comment to it?

ilyasoifer (approved):

PR #18, opened 2023-02-06
[BIOIN-803] Script to clean DeepVariant gVCF before joint calling 

ilyasoifer:


ugvc/joint/cleanup_gvcf_before_calling.py
```diff
@@ -0,0 +1,95 @@
...
+                    else:  # (tmp.samples[0]['GT']==(None,None)):
+                        count_skipped += 1
+                        continue
+                    buffer_chrom = ""
+                    buffer_span = 0
+                called_var_in_buffer = False
+            # now for the new deletion
+            dels_in_rec = pu.is_deletion(rec)
+            lens_in_rec = pu.indel_length(rec)
+            del_lens = [lens_in_rec[i] for i, x in enumerate(dels_in_rec) if x] + [0]
```
gilhornung: Instead of using the list, do the max operation here

gilhornung (approved):
Minor comment

ugvc/joint/cleanup_gvcf_before_calling.py
ilyasoifer: done

PR #19, opened 2023-02-06
Doronst/bio 896/report improvements

doron-st:
Mostly refactoring and change to vcfeval metrics
Report cotent changes:
* Calculate TP/FP/FN from base/call columns
* Calculate max_recall by reducing only (base=FN, call=NA) from total true variants
* Added "# initial fn" to performance dataframes
* Added option for model_name in input_params
* Refactoring into reusable functions
* Add verbosity parameter that controls report detail level


doron-st:
I added the high h-mers, perhaps not right for this PR.
Can revert if we want

.github/workflows/python-package-conda.yml
```diff
@@ -61,5 +61,6 @@ jobs:
       - name: Test with pytest
         run: |
           conda activate genomics.py3
+          export PYTHONPATH=$(pwd)
```
ilyasoifer: Why is this needed?

test/unit/comparison/test_vcf_pipeline_utils.py
```diff
@@ -26,7 +27,10 @@ def test_fix_errors():
     )
 
     # (None, TP) (None,FN_CA)
-    assert df[(df["call"].isna()) & ((df["base"] == "TP") | (df["base"] == "FN_CA"))].size == 0
+    pd.set_option("display.max_columns", None)
+    take = df[(df["call"].isna()) & ((df["base"] == "TP") | (df["base"] == "FN_CA"))]
+    print(take)
```
ilyasoifer: Print - might be my fault, sorry

ugvc/comparison/comparison_pipeline.py
```diff
@@ -8,181 +8,164 @@
...
+            self.call_sample,
+            self.truth_sample,
+            self.ignore_filter,
         )
-        output_prefix = f"{output_prefix}.vcfeval_concordance"
-    else:
-        vpu.run_genotype_concordance(
+        annotated_concordance_vcf = self.vpu.annotate_tandem_repeats(concordance_vcf, self.ref_genome)
+
+        high_conf_calls_vcf = self.vpu.filter_bad_areas(
```
ilyasoifer: I think we can remove the option of deleting the runs (left from the past)

ugvc/comparison/vcf_pipeline_utils.py
```diff
@@ -19,27 +21,47 @@
 
 
 class VcfPipelineUtils:
+
```
ilyasoifer: Empty line before docs and docs are not quite informative (probably also my fault somehow)

ugvc/comparison/vcf_pipeline_utils.py
```diff
@@ -50,37 +72,44 @@ def combine_vcf(self, n_parts: int, input_prefix: str, output_fname: str):
         self.index_vcf(output_fname)
 
     def index_vcf(self, vcf: str):
-        """Tabix index on VCF"""
+        """Tabix index on VCF
+
+        Parameters
+        ----------
+        vcf : str
+            Description
```
ilyasoifer:
```suggestion
            Input VCF.gz file
```

ugvc/comparison/vcf_pipeline_utils.py
```diff
@@ -207,22 +206,106 @@ def run_vcfeval_concordance(
         self.fix_vcf_format(os.path.join(vcfeval_output_dir, "output"))
 
         # make the vcfeval output file without weird variants
```
ilyasoifer, line 207:
```suggestion
        # make the vcfeval output file where we collapse identical positions into multiallelic variants merging the BASE/CALL fields
```

ugvc/comparison/vcf_pipeline_utils.py
```diff
@@ -312,7 +404,18 @@ def transform_hom_calls_to_het_calls(self, input_file_calls: str, output_file_ca
 
 
 def _fix_errors(df):
-    """Fixes errors that complicated VCFEVAL VCF format introduces"""
+    """Fixes errors that complicated VCFEVAL VCF format introduces
```
ilyasoifer: I will update the docs here and in other places

ugvc/comparison/vcf_pipeline_utils.py
```diff
@@ -342,6 +445,20 @@ def _fix_errors(df):
...
+        Description
+    concordance_format : str
+        Description
+
+    Returns
+    -------
+    defaultdict
+        Description
+    """
     call_sample_ind = 1 if concordance_format == "VCFEVAL" else 0
```
ilyasoifer: Do we need concordance_format == "VCFEVAL" at all?

ugvc/comparison/vcf_pipeline_utils.py
```diff
@@ -393,71 +508,62 @@ def vcf2concordance(
...
+    concordance_df["gt_ground_truth"] = concordance_df["gt_ground_truth"].map(
+        lambda x: (None, None) if x == (None,) else x
+    )
+
     concordance_df["indel"] = concordance_df["alleles"].apply(lambda x: len({len(y) for y in x}) > 1)
 
-    if concordance_format == "VCFEVAL":
-        concordance_df = _fix_errors(concordance_df)
+    pd.set_option("display.max_columns", None)
+    print("vcf2concordance: after initial construction")
```
ilyasoifer: Prints?

ilyasoifer (commented):
Hi looks very nice, some comments, but they are minor 
Suggestions:
1. Switch to papermill 
2. Modify the extended report too

.github/workflows/python-package-conda.yml
doron-st: So that jupyter will be able to import ugvc

test/unit/comparison/test_vcf_pipeline_utils.py
doron-st: right

ugvc/comparison/vcf_pipeline_utils.py
doron-st: nope

doron-st:
ok, I'm done with review changes.
Want to finish it off? (docs and the hmer_runs thing)

BTW this miserable merge I did in the middle prevents from squashing...

doron-st:
> Hi looks very nice, some comments, but they are minor Suggestions:
> 
> 1. Switch to papermill
> 2. Modify the extended report too

papermill and more stuff out the window in a separate PR

ilyasoifer:
Fun, on it, will be done by the evening

.github/workflows/python-package-conda.yml
ilyasoifer: It does not quite make sense, ugvc is installed in the environment

doron-st: 1. Where is it installed? I don't see it installed in the workflow.
    It is added to the path by pytest, but this is not enough for jupyter.
    
    Here you can see the test passes (since I added ugvc to path with PYTHONPATH), and later it fail after I remove it from path:
    
image

2. Another question is why do we need to export PYTHONPATH when running regular jupyter convert with installed ugvc. When I think of installed I think of "cd VariantCalling; pip install ."

Here you can see the failure of the previous jupyter convert command, and then when I add to path it completes:
image

doron-st: I have just checked, and papermill doesn't solve this

ilyasoifer: I miss something? On my laptop I don't need this
(genomics.py3) ilyas@ilya-laptop:~/proj/BioinfoResearch/VariantCalling$ echo $PYTHONPATH


(genomics.py3) ilyas@ilya-laptop:~/proj/BioinfoResearch/VariantCalling$ pytest test/system/test_vc_report.py 
=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.7.8, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /home/ilyas/proj/BioinfoResearch/VariantCalling
plugins: cov-3.0.0, mock-3.6.1, xdist-2.3.0, forked-1.3.0
collected 4 items                                                                                                                                                                                                 

test/system/test_vc_report.py ..

ilyasoifer: oh - maybe __init__.py was missing in the reports?

ilyasoifer: ok, fixed

ilyasoifer (approved):

PR #37, opened 2023-05-31
genome-wize_normalize_reads_count

TammushB:


ilyasoifer:
@TammushB - i am waiting until the checks pass, right?

TammushB:
@ilyasoifer I think we are done here. we added a normalization script that includes the option of gender estimation. also we updated cn.mops pull version to include latest fixes. 
please review.

ugvc/cnv/normalize_reads_count.R
```diff
@@ -0,0 +1,314 @@
...
+#' @param quSizeFactor Quantile of the sizeFactor if sizeFactor is set to "quant".
+#' 0.75 corresponds to "upper quartile normalization". Real value between 0 and 1. Default = 0.75.
+#' @param ploidy An integer value for each sample or each column in the read
+#' count matrix. At least two samples must have a ploidy of 2. Default = "missing".
+#' @examples
+#' data(cn.mops)
+#' X.norm <- normalizeChromosomes(X)
+#' @return A data matrix of normalized read counts with the same dimensions
+#' as the input matrix X.
+#' @author Guenter Klambauer \email{klambauer@@bioinf.jku.at}
```
ilyasoifer: Better to refer to the original code and add yourself as an author

ugvc/cnv/normalize_reads_count.R
```diff
@@ -0,0 +1,314 @@
...
+
+		if (any(is.na(globalsizeFactors))) stop("NAs in readcount matrix.")
+		if (any(globalsizeFactors==0)) stop("Zero columns in readcount matrix.")
+
+		if (ncol(X)==1){
+			Y <- X
+		} else {
+
+			Y <- matrix(0,nrow=nrow(X),ncol=ncol(X))
+			chrIdx <- which((chr != "chrX") & (chr != "chrY"))
```
ilyasoifer: Need some safeguard for sex chromosome namings

ugvc/cnv/normalize_reads_count.R
```diff
@@ -0,0 +1,314 @@
...
+out_prefix <- args$out_prefix
+ploidy_file <-args$ploidy
+
+cohort_reads_count<- readRDS(file = cohort_reads_count_file )
+if (!is.null(ploidy_file)) {
+	ploidy_vector<-readLines(ploidy_file)
+	cohort_reads_count_normalized <- normalizeChromosomesGenomewize(cohort_reads_count,ploidy=ploidy_vector)
+	saveRDS(cohort_reads_count_normalized,paste(out_prefix,"cohort_reads_count.norm.rds",sep=""))
+} else {
+	print("Gender will be estimated by chrX coverage")
```
ilyasoifer:
```suggestion
	print("Sex will be estimated by chrX coverage")
```

ugvc/cnv/normalize_reads_count.R
```diff
@@ -0,0 +1,314 @@
...
+					Ytmp <- X %*% diag(correctwiththis)
+					Ytmp[idxSG, ] <- 0
+
+					Y <- Ytmp
+				}
+
+		} # over chr
+
+	#handle Sex chromosomes(chrX,chrY) by Gender ploidy value.
+	#for male we will multiply normalized X and Y coverage with 2.
```
ilyasoifer, line 217: Why do we need to multiply Y coverage by 2?

ugvc/cnv/normalize_reads_count.R
```diff
@@ -0,0 +1,314 @@
...
+  right_dist = X[X>in_cutoff]
+  left_variance = var(left_dist)
+  right_variance = var(right_dist)
+  sum_var = left_variance+right_variance
+  return(sum_var)
+}
+
+#' @title whole Genome-wize Normalization of NGS data.
+#'
+#' @description Normalize quantitative NGS data in order to make counts comparable over
```
ilyasoifer, line 50: Why don't we add this function to cn.mops actually?

ilyasoifer (changes requested):
Hi look good. 
I think that the new normalization function should live in cn.mops repo

ugvc/cnv/normalize_reads_count.R
TammushB: The main reason for not moving this script to cn.mops repo is that it does not have tests within it. 
We have a implemented a test for this particular script within current repo and it will be useful for further debugging.

ilyasoifer: Got it... sounds like we made something pretty ugly. You can merge it for now. Can you please leave jira issue to fix this? 
I think we should create a conda package for our version of cn.mops probably to make it more reasonable

TammushB: Agreed: https://ultimagen.atlassian.net/browse/BIOIN-1121

ugvc/cnv/normalize_reads_count.R
TammushB: in current implementation we need to do this in order to not read chrY as deletion for male samples.

ilyasoifer: I am sorry, I don't understand this. The Y chromosomes are all the same in all males, no? I understand why you would multiply chrX by two in males

ugvc/cnv/normalize_reads_count.R
```diff
@@ -0,0 +1,315 @@
...
+#' X.norm <- normalizeChromosomes(X)
+#' @return A data matrix of normalized read counts with the same dimensions
+#' as the input matrix X.
+#' @author Tammy Biniashvili \email{tammy.biniashvili@ultimagen.com}.
+#' original script source is cn.mops repository: https://github.com/bioinf-jku/cn.mops/blob/f9fef3f2c6a5a0bd323a754a0256b840565401ba/R/normalizeChromosomes.R
+#' @export
+
+
+normalizeChromosomesGenomewize <- function(X, chr, normType="poisson", sizeFactor="mean",
+		qu=0.25, quSizeFactor=0.75, ploidy, chr_X = "chrX", chr_Y="chrY"){
```
ilyasoifer:
```suggestion
		qu=0.25, quSizeFactor=0.75, ploidy, chr_X_name = "chrX", chr_Y_name="chrY"){
```

ilyasoifer (approved):
Looks good - just two minor comments

ugvc/cnv/normalize_reads_count.R
TammushB: Since normalization is done Genome-wise, if we will not multiply it  by 2 we will get CN=1 for that chromosome same as for chrX. in our next update, we will try to run chrY separately from the cohort, and then we will not have to multiply by2.

PR #39, opened 2023-06-18
[BIOIN-959] upgrade  genomics.py3

ilyasoifer:
Hi, 
This is both an FYI and a request for review. 
I spent some time upgrading our four-year-old conda environment to the new python (3.8) and the new Java (17). 
The upgrade updated samtools, bcftools, gatk and picard so that the installed versions contains almost all our updates. 
Unfortunately, due to some API changes in sklearn, numpy and pandas I had to make some changes of the code. I think I fixed the issues that were covered by the tests, but obviously some things could have broken. 

After this merge, I strongly suggest removing and re-creating the conda environment to continue using VariantCalling.

Specifically - these were the two main issues that you might want to consider checking in your code. 
1. pandas.append was deprecated
2. dtypes np.bool. np.int and np.float were replaced  by bool, int and float

Other than that, things seem to be working fine
Looking forward for feedback, please let me know if you were using some package that got removed.


setup/environment.yml
```diff
@@ -1,428 +1,479 @@
...
+      - mpld3==0.5.9
+      - pdf2image==1.16.3
+      - pymupdf==1.22.5
+      - pypdf2==3.0.1
+      - reportlab==4.0.4
+      - sigprofilerassignment==0.0.30
+      - sigprofilermatrixgenerator==1.2.17
+      - sigprofilerplotting==1.3.14
+      - simppl==1.0.7
+prefix: /home/ilyas/anaconda3/envs/genomics.py3
```
kenissur: Might be best to delete this last line

https://stackoverflow.com/questions/41274007/anaconda-export-environment-file

kenissur (commented):
Thanks Ilya!

Say, why not go to the newest python version?

ilyasoifer:
> Thanks Ilya!
> 
> Say, why not go to the newest python version?

I will try, uff

ilyasoifer:
> > Thanks Ilya!
> > Say, why not go to the newest python version?
> 
> I will try, uff

@kenissur @mayalevy Spent a few hours on that and could not make this happen. There is some unclear conflict between python 3.10, picard3 and R that I can't figure out. One day someone stronger than me will solve this, but at least for now we will be able to run a current GATK from the environment

setup/environment.yml
ilyasoifer: Fixed

doron-st (approved):
Not sure how to review this :)

ugvc/comparison/vcf_pipeline_utils.py
```diff
@@ -395,7 +395,9 @@ def _fix_errors(df: pd.DataFrame) -> pd.DataFrame:
     # (FP_CA,FN_CA), (FP_CA,None) - Fake a genotype from ultima such that one of the alleles is the same (and only one)
     df.loc[(df["call"] == "FP_CA") & ((df["base"] == "FN_CA") | (df["base"].isna())), "gt_ground_truth"] = df[
         (df["call"] == "FP_CA") & ((df["base"] == "FN_CA") | (df["base"].isna()))
-    ]["gt_ultima"].apply(lambda x: ((x[0], x[0]) if (x[1] == 0) else ((x[1], x[1]) if (x[0] == 0) else (x[0], 0))))
+    ]["gt_ultima"].apply(
+        lambda x: ((x[0], x[0]) if (len(x) < 2 or (x[1] == 0)) else ((x[1], x[1]) if (x[0] == 0) else (x[0], 0)))
+    )
     return df
```
gilhornung, line 401: Why did you add len(x) < 2?
Also - how is this related to geniomics.py3

gilhornung (approved):
עשית פה עבודת נמלים.

ugvc/comparison/vcf_pipeline_utils.py
ilyasoifer: Behaviour of `bcftools norm` changed slightly in this new version and this had a weird effect downstream

PR #41, opened 2023-06-21
Report fixes

GatKrieger:
Present the sample's basename and h5 path, and also display edits


mayalevy (approved):
I think it is fine, just run the nightly tests that have failed

PR #43, opened 2023-06-28
use qual_af clusters as labels

doron-st:
Closes the loop with qual/af clustering -> lib-prep-recalibration -> combine scores
image

### DIFF
--- a/ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
+++ b/ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
@@ -1,5 +1,7 @@
+import gc
 import os
 
+import pandas as pd
 from simppl.cli import get_parser, get_simple_pipeline
 
 from ugvc.pipelines.mrd import featuremap_to_dataframe
@@ -31,6 +33,7 @@ def run(argv):
     args = parser.parse_args(argv[1:])
     print(args)
 
+    nb_dir = args.notebooks_dir
     out_dir = args.out_dir
     os.makedirs(out_dir, exist_ok=True)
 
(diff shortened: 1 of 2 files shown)


ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
```diff
@@ -31,49 +33,103 @@ def run(argv):
...
         )
         sp.print_and_run_clt(
-            featuremap_to_dataframe.run, [], {"-i": vcf_file, "-r": args.ref_fasta, "-f": "TGCA", "-o": parquet_file}
+            featuremap_to_dataframe.run,
+            [],
+            {"-i": featuremap_on_calls_vcf, "-r": args.ref_fasta, "-f": "TGCA", "-o": featuremap_on_calls_parquet},
+        )
+        qual_af_df = pd.read_parquet(qual_af_clusters_on_calls)
+        featuremap_on_calls_df = pd.read_parquet(featuremap_on_calls_parquet)
+        featuremap_on_calls_df = featuremap_on_calls_df.drop(columns=["qual", "filter"])
```
kenissur: You can specify which columns you want read from the parquet file instead of reading and dropping.

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
```diff
@@ -31,49 +33,103 @@ def run(argv):
...
+        qual_af_df.merge(featuremap_on_calls_df, on=["chrom", "pos", "ref", "alt"]).to_parquet(labeled_training_set)
```
kenissur, line 97: Does that work in a reasonable time? I would worry about a merge on 4 columns. Setting as index and running "join" is much faster than "merge" on columns in my experience, and I think you can safely drop the "ref". Maybe even the "chrom"?

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
```diff
@@ -31,49 +33,103 @@ def run(argv):
...
+        del qual_af_df
+        del featuremap_on_calls_df
+        gc.collect()
```
kenissur, line 100: Interesting, I never an explicit garbage collector call. That is all to clear up memory, right?

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
kenissur: BTW, if you do set an index with multiple columns it's best to reset it before saving

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
```diff
@@ -31,49 +33,103 @@ def run(argv):
...
+            )
+            sp.print_and_run_clt(
+                featuremap_to_dataframe.run,
+                [],
+                {"-i": vcf_file, "-r": args.ref_fasta, "-f": "TGCA", "-o": parquet_file},
+            )
+        fp_df = pd.read_parquet(fp_args[2])
+        tp_df = pd.read_parquet(tp_args[2])
+        fp_df["label"] = False
+        tp_df["label"] = True
```
kenissur: Doesn't it warn you to do this instead?
```suggestion
        tp_df.loc[:, "label"] = True
```

kenissur (approved):
Looks good as always, I reviewed as best as I could. The clustering part is done on an external notebook that this code gets as input?

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
```diff
@@ -31,49 +33,103 @@ def run(argv):
...
     if args.calls_vcf:
         calls_vcf = f"{out_dir}/calls.snps.pass.vcf.gz"
-        featuremap_on_calls = f"{out_dir}/featuremap_on_calls.vcf.gz"
+        calls_parquet = f"{out_dir}/calls.snps.pass.parquet"
+        featuremap_on_calls_vcf = f"{out_dir}/featuremap_on_calls.vcf.gz"
+        featuremap_on_calls_parquet = f"{out_dir}/featuremap_on_calls.parquet"
+        qual_af_clusters_on_calls = f"{out_dir}/qual_af_clusters_on_calls.parquet"
         sp.print_and_run(f"bcftools view --type snps -f PASS -m2 -M2 {unfiltered_calls} -Oz -o {calls_vcf}")
         sp.print_and_run(
             f"bedtools intersect -a {featuremap_vcf} -b {calls_vcf} -header | "
```
kenissur: Does "bedtools intersect" make sure the alts in both vcfs match?

doron-st: Of course it doesn't.
I am planning to refine this

doron-st:
> Looks good as always, I reviewed as best as I could. The clustering part is done on an external notebook that this code gets as input?

Thanks. Yes. The notebooks are still in bioinfo for now

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
doron-st: Well, It is easier to do it in the negative direction. I know which columns I don't want

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
doron-st: Yes, it is quite fast. 
We are only looking at pass-filter loci here.
I haven't noticed any problem with the index of the labeled_training_set, but will look again in the analysis that uses it.

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
doron-st: Indeed

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
doron-st: Probably does, I haven't ran this use-case actually. will fix.

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
doron-st: The index looks fine to me:
image

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
doron-st: Actually it makes no difference at all, since we later use the df merge to select the alleles.
This is just preliminary filtering for efficiency.

However I do classify variants as true/false in the end by position (if we have GTR), which should be refined.

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
kenissur: Sure. Maybe check that they exist before dropping as a precaution.

ugvc/pipelines/lpr/train_lib_prep_recalibration_model.py
kenissur: I meant that if you use a multiindex like I suggested it's best not to save this way, nvm

PR #45, opened 2023-07-05
Rinas/bqsr

rinashalibo:


ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+import pysam
+from joblib import Parallel, delayed
+from tqdm import tqdm
+
+from ugvc.dna.utils import revcomp
+from ugvc.utils.cloud_sync import cloud_sync
+from ugvc.utils.misc_utils import set_pyplot_defaults
+from ugvc.vcfbed.variant_annotation import get_cycle_skip_dataframe, get_motif_around
+
+set_pyplot_defaults()
```
doron-st: doing this in global scope can affect other unrelated analyses, right?

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def iterate_vcf_in_chunks(vcf_file, chunk_size, model, output_file, chrstr, params, show_progress=True):
```
doron-st: Can be private? (__foo).
Also I would rename it to be more indicative to what the function does. e.g. __bqsr_inference_per_chromosome

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+            '##INFO=<ID=ML_QUAL,Number=1,Type=Float,Description="Quality assigned using an ML model">'
+        )
+
+        # Calculate the total number of chunks
+        # total_variants = vcf_in.count_records()
+        # total_chunks = (total_variants - 1) // chunk_size + 1
+
+        # Iterate over the variants in chunks
+        variants = vcf_in.fetch(chrstr)
+        iterator = tqdm(total=None, disable=not show_progress)
```
doron-st: rename to progress_tracker? something less generic.

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+        chunk = [None] * chunk_size
```
doron-st: why don't we use a python list for chunk?

ugvc/mrd/bqsr_inference_utils.py
doron-st: Also, I see you pass many arguments, which makes it likely that this will better be a private method inside a class.
The class will hold all the common arguments as fields.

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+
+        # Process the remaining variants (if any) in the last chunk
+        if index > 0:
+            process_chunk(chunk[:index], model, vcf_out, params)
+            iterator.update(1)
+
+        iterator.close()
+
+
+def process_chunk(chunk, model, vcf_out, params):
```
doron-st: also would better be a private method

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+
+    # Convert variants to a pandas DataFrame
```
doron-st: why don't we use featuremap_to_dataframe?
The next lines up to ~135 look like duplication

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+    )
+    _ = cloud_sync(
+        "gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.fasta.fai",
+        local_dir_in="/data1",
+    )
+
+    def parallel_iterate_over_chunks(chrstr):
+        # global params
+
+        vcf_file = cloud_sync(params["featuremap_file"], local_dir_in="/data1")
```
doron-st: If this script is to be run within WDL framework we shouldn't worry about file localization, don't we?
Can assume the files are found locally

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+        model = joblib.load(params["model_path"])
+
+        vcf_out = pjoin(params["workdir"], params["out_basename"] + f"_output_{chrstr}.vcf.gz")
+
+        iterate_vcf_in_chunks(vcf_file, chunk_size, model, vcf_out, chrstr, params)
+
+    # inference on input featuremap
+    chrlist = []
+    for c in np.arange(1, 23).astype(str):
+        chrlist.append("chr" + c)
```
doron-st: I'm a bit concerned about how this is tied specifically to hg38 format.
It can wait for future version

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,363 @@
...
+import xgboost as xgb
+from joblib import dump
+from matplotlib import pyplot as plt
+from sklearn.model_selection import train_test_split
+
+from ugvc.dna.format import DEFAULT_FLOW_ORDER
+from ugvc.mrd.mrd_utils import featuremap_to_dataframe
+from ugvc.utils.misc_utils import set_pyplot_defaults
+
+set_pyplot_defaults()
```
doron-st: same here

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,363 @@
...
+    if not isfile(f_out):
+        # FP file - single substitution
+        print("prepare FP vcf file...")
+        f_fp = single_substitution_file
+        f_fp_training = pjoin(params["workdir"], f"{params['data_name']}.fp.vcf.gz")
+
+        if not isfile(f_fp_training):
+            commandstr = f"bcftools view {f_fp} {params['training_chrom']} -o {f_fp_training} -O z"
+            print(commandstr)
+            subprocess.call(commandstr, shell=True)
```
doron-st: I usually prefer using simple_pipeline over subprocess.

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,363 @@
...
+
+    with open(f"{params['workdir']}/params.json", "w", encoding="utf-8") as f:
+        json.dump(params, f)
+
+
+def get_precision(df, label):
+    return (df["label"] == label).sum() / df.shape[0] if df.shape[0] > 0 else np.nan
+
+
+def get_recall(df, df_before_filter, label):
```
doron-st: please update to the latest functions I've rewritten.
As we said should go in stats_utils or metrics_utils

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,363 @@
...
+    return (
+        (df["label"] == label).sum() / (df_before_filter["label"] == label).sum()
+        if (df_before_filter["label"] == label).sum() > 0
+        else np.nan
+    )
+
+
+def get_precision_recall(df, max_score, label, score="qual", cumulative=False, apply_log_trans=True):
```
doron-st: please update

ugvc/reports/bqsr_train_report.ipynb
doron-st, line 1: Where do you call the report?

ugvc/reports/bqsr_train_report.ipynb
```diff
@@ -0,0 +1,811 @@
...
...Y2JjmMzlsXfp7OKz3WJZnTXf0ZYA2wrPuzco9VK2nKGKYkjSvJUd03qc/rvnrojiRHJFnX/fzVJFcnefuIzX4vycbuzz/u7ufnknyju69vJFk8Sfs7ktye5LtJbui+Fupc4J1JNid55zj1zk/ytSR/leS/JflBkgVJliS5fUS/301yTvfzbya5pXusLyd5fi9jU1V3VtVdY6xaBVxaVX9fVfcBW+m8UPWVwIuq6ubqPC3588Cv9nIsSYPNMCVpXFV1C533e/0B8DHgC8ChwBLgNcBpwNGjNvvbqloOfJrOK53ofv58VR0BfBE4f5L2DwJvrqrXAm+rqp922y6rqiOr6rJxSj4buLGqfqFb9+Ie/piXV9VR3WPdCbynh20mchDw4Ijl4W7bQd3Po9slzXCGKUmTORd4IzBEJ1AdC/z3qnqqqv4fcN2o/peM+L0raB0NXNz9/OfdfUzU/pfARUl+k857tHp1HJ3AR1VdBfxND9scnuQvkmwBTqZzea7FWPdB1QTtkmY4w5SkybwMeCFwADCPsUPBSDXO5/H6PKu9qk4H/jNwMLC5+8LWXo2175088++7eSM+XwScWVWvAT40at3uGKZT9y6LgG3d9kVjtEua4QxTkiazATiLzmW4jwI3Av8qyT5JXgGsGNX/nSN+39z9fBOdm7KhM/tz40TtSQ6pqm9X1QeBR+iEk8foBLqJ3NDdD0lOAF7abf8h8PLuPVXPA/7FiG0OAB5KMnfXto2uBE5K8rwkS+ncaL6x+xb6x5K8vvstvncBV/TheJKm2b7TXYCkwZXkXcDOqrq4+420m4DL6cyy3A7cDXwb+PGIzZ6X5Nt0/rG2utv274ELk/wesB04dZL285IsozML9g3gu8ADwNokm4E/HOe+qQ8BlyT5DvDN7jZU1c+SnNut9T7g+yO2Oavb/gNgC5MHtl1jcyLwx8BC4Kokm6vqzVV1R5IvAd+jMyN2RlU92d3st+jMhO0PXNP9kTTDpfOlEknqXZIXVtXj3ctvG4FjuvdPDZQk9wNDVfXIdNciafZyZkrS7vhqkpcA+wEfHsQgJUl7ijNTkmacJKcC7xvV/JdVdcYUHOsC4JhRzZ+qqj/t97EkzUyGKUmSpAZ+m0+SJKmBYUqSJKmBYUqSJKmBYUqSJKmBYUqSJKnB/wfNmEKcRUcUHgAAAABJRU5ErkJggg==",
```
doron-st: please clear outputs and re-save before commiting

ugvc/pipelines/mrd/bqsr_inference.py
```diff
@@ -0,0 +1,97 @@
...
+        help="""training chromosome""",
+    )
+    return parser.parse_args(argv[1:])
+
+
+def run(argv: list[str]):
+    """BQSR inference: load model, run inference and write quality for all reads in the input featuremap"""
+    args = parse_args(argv)
+
+    bqsr_inference(
```
doron-st: I would put the main function (bqsr_inference) in this file.
It will be simpler to understand the pipeline like this.

ugvc/pipelines/mrd/bqsr_train.py
```diff
@@ -0,0 +1,81 @@
...
+        help="""flow order - required for cycle skip annotation """,
+    )
+    return parser.parse_args(argv[1:])
+
+
+def run(argv: list[str]):
+    """BQSR: Train xgb model, save model and data for report"""
+    args = parse_args(argv)
+
+    bqsr_train(
```
doron-st: same for bqsr_train, I would pur here the code

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,363 @@
...
+        max_depth=params["max_depth"],
+        n_jobs=params["n_jobs"],
+        enable_categorical=True,
+    )
+    xgb_classifier.fit(X_train, y_train["label"])
+
+    # save classifier and data, generate plots for report
+    save_model_and_data(xgb_classifier, X_test, y_test, params)
+    # generate plots
+    # generate_plots(xgb_classifier,X_test,y_test)
```
doron-st: Don't we want to run the report here?

doron-st (commented):
Very nice! Good to see how it comes alive.
See comments.

doron-st:
General comment:
In ugvc standards, we strive to annotate the types of function arguments, and add a docstring for each function.

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+from ugvc.dna.utils import revcomp
+from ugvc.utils.cloud_sync import cloud_sync
+from ugvc.utils.misc_utils import set_pyplot_defaults
+from ugvc.vcfbed.variant_annotation import get_cycle_skip_dataframe, get_motif_around
+
+set_pyplot_defaults()
+
+
+def iterate_vcf_in_chunks(vcf_file, chunk_size, model, output_file, chrstr, params, show_progress=True):
+    print(f"fetching chr {chrstr} from {vcf_file}\n")
```
kenissur: As we discussed, let's use a bed file instead of a specific chromosome

ugvc/mrd/bqsr_inference_utils.py
kenissur: This is a preallocated list of size chunk_size with None values, why is it not good?

ugvc/mrd/bqsr_inference_utils.py
kenissur: I would also rename to something less generic

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+    params["out_basename"] = out_basename
+    params["featuremap_path"] = featuremap_path
+    params["params_path"] = params_path
+    params["model_path"] = model_path
+    params["chunk_size"] = 20000  # vcf is loaded in chunks, this number is the amount of reads in chunk
+
+    params["ref_fasta"] = cloud_sync(
+        "gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.fasta",
+        local_dir_in="/data1",
+    )
```
kenissur: Shold be an input, not hard coded, and should not use cloud_sync

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+    if len(out_basename) > 0 and not out_basename.endswith("."):
+        out_basename += "."
+
+    params = {}
+    params["workdir"] = out_path
+    params["out_basename"] = out_basename
+    params["featuremap_path"] = featuremap_path
+    params["params_path"] = params_path
+    params["model_path"] = model_path
+    params["chunk_size"] = 20000  # vcf is loaded in chunks, this number is the amount of reads in chunk
```
kenissur: Should be an input

ugvc/mrd/bqsr_inference_utils.py
kenissur: +1

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+    out_path: str,
+    out_basename: str = "",
+):
+
+    # init
+    os.makedirs(out_path, exist_ok=True)
+    if len(out_basename) > 0 and not out_basename.endswith("."):
+        out_basename += "."
+
+    params = {}
```
kenissur: I don't understand the purpose of this dictionary, I would either keep these as separate variables or create a class like Doron suggested.

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+from ugvc.utils.misc_utils import set_pyplot_defaults
+from ugvc.vcfbed.variant_annotation import get_cycle_skip_dataframe, get_motif_around
+
+set_pyplot_defaults()
+
+
+def iterate_vcf_in_chunks(vcf_file, chunk_size, model, output_file, chrstr, params, show_progress=True):
+    print(f"fetching chr {chrstr} from {vcf_file}\n")
+    with pysam.VariantFile(vcf_file) as vcf_in, pysam.VariantFile(output_file, "w", header=vcf_in.header) as vcf_out:
+
```
kenissur: Add this to get the list of contigs (chromosomes) to run inference on so you don't have to hard code them:
contigs = [x.name for x in header.contigs.values() if "_" not in x.name and "-" not in x.name and x.name != "chrM" and x.name != "chrEBV"]

Then pass it on to the inference function.

We could remove the if statements to make it more general, this way the inference will also run on contigs where there are no featuremap entries so you would just have to see it doesn't cause issues.

ugvc/mrd/bqsr_inference_utils.py
kenissur: We definitely should, and seeing this now I realize the balanced ePCR tags should be handled in one dedicated place. This is something we should discuss because there are hanging changes in another branch.

ugvc/mrd/bqsr_inference_utils.py
kenissur: Please see my earlier comment - it's easy to extract from the vcf header that you already have.

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+    # inference on input featuremap
+    chrlist = []
+    for c in np.arange(1, 23).astype(str):
+        chrlist.append("chr" + c)
+
+    # Parallel loop on chromosomes and then unite results
+    Parallel(n_jobs=-1)(delayed(parallel_iterate_over_chunks)(chr_name) for chr_name in chrlist)
+    united_vcf_file = pjoin(params["workdir"], f"{params['out_basename']}.featuremap.ML_qual.vcf.gz")
+    commandstr = f"bcftools concat {params['workdir']}/*featuremap_output_chr*.vcf.gz -o f{united_vcf_file} -O z"
+    print(commandstr)
```
kenissur: Should either be removed or within a logger object

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+
+    # inference on input featuremap
+    chrlist = []
+    for c in np.arange(1, 23).astype(str):
+        chrlist.append("chr" + c)
+
+    # Parallel loop on chromosomes and then unite results
+    Parallel(n_jobs=-1)(delayed(parallel_iterate_over_chunks)(chr_name) for chr_name in chrlist)
+    united_vcf_file = pjoin(params["workdir"], f"{params['out_basename']}.featuremap.ML_qual.vcf.gz")
+    commandstr = f"bcftools concat {params['workdir']}/*featuremap_output_chr*.vcf.gz -o f{united_vcf_file} -O z"
```
kenissur: Is the output guaranteed to be sorted this way?

ugvc/mrd/bqsr_inference_utils.py
kenissur: or within a simppl command like Doron does

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+
+    features = features.astype({col: "category" for col in categorical_features})
+
+    # scaled_features = preprocessing.scale(features)  # Scale features if needed
+    predicted_qualities = model.predict_proba(features)
+
+    # Assign the new quality value to each variant and write to the output file
+    for i, variant in enumerate(chunk):
+        # Add a new INFO field to the variant indicating the ML quality value
+        # variant.info.__setitem__("ML_QUAL", -10 * np.log10(predicted_qualities[i][0]))
```
kenissur: remove commented line

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+    # # ['cycle_skip_status', 'ref_alt_motif', 'left_motif', 'right_motif',
+    # #   'X_SCORE', 'X_EDIST', 'X_LENGTH', 'X_MAPQ', 'X_INDEX', 'is_mixed']
+
+    features = df[model.feature_names_in_]
+    motif_only_features = ["ref_alt_motif", "left_motif", "right_motif"]
+    categorical_features = ["cycle_skip_status"] + motif_only_features
+
+    features = features.astype({col: "category" for col in categorical_features})
+
+    # scaled_features = preprocessing.scale(features)  # Scale features if needed
```
kenissur: remove commented line

ugvc/mrd/bqsr_inference_utils.py
```diff
@@ -0,0 +1,205 @@
...
+    df = df.assign(is_cycle_skip=df["cycle_skip_status"] == "cycle-skip")
+
+    df.loc[:, "ref_alt_bases"] = df["ref"].str.cat(df["alt"], sep="")
+    df.loc[:, "ref_alt_motif"] = df["ref_motif"].str.cat(df["alt"], sep="")
+    df.loc[:, "left_motif"] = df["left_motif"].str.slice(start=0, stop=3)
+    df.loc[:, "right_motif"] = df["right_motif"].str.slice(start=1, stop=4)
+
+    # # ['cycle_skip_status', 'ref_alt_motif', 'left_motif', 'right_motif',
+    # #   'X_SCORE', 'X_EDIST', 'X_LENGTH', 'X_MAPQ', 'X_INDEX', 'is_mixed']
+
```
kenissur: remove commented line

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+from ugvc.mrd.mrd_utils import featuremap_to_dataframe
+from ugvc.utils.misc_utils import set_pyplot_defaults
+
+set_pyplot_defaults()
+
+
+def model_params(params):
+
+    params["balance_motifs"] = True
+    params["eval_only"] = False
```
kenissur: This is not used. One of the reasons that I think this dictionary structure is not good is that the interpreter can't figure that out this way.

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+from ugvc.vcfbed.variant_annotation import get_cycle_skip_dataframe, get_motif_around
+from ugvc.dna.format import DEFAULT_FLOW_ORDER
+from ugvc.mrd.mrd_utils import featuremap_to_dataframe
+from ugvc.utils.misc_utils import set_pyplot_defaults
+
+set_pyplot_defaults()
+
+
+def model_params(params):
+
```
kenissur: I think this function should be removed, see suggestions below.

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+        "X_MAPQ",
+        "X_INDEX",
+        "max_softclip_len",
+        "X_FC1",
+    ]
+
+    features = categorical_features + non_categorical_features
+    if "is_mixed" in data.columns:
+        features.append("is_mixed")
+
```
kenissur, line 537: all of the features should be input parameters.

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+    for col in categorical_features:
+        data[col] = data[col].astype("category")
+
+    data = balance_motifs(data, params)
+    data["label"][data["label"] == "TP"] = 1
+    data["label"][data["label"] == "FP"] = 0
+    data = data.astype({"label": "int16"})
+
+    X_train, X_test, y_train, y_test = train_test_split(
+        data[features], data[label_columns], test_size=0.33, random_state=42
```
kenissur: I suggest making test_size a numerical value and not a ratio. IMO you don't generally need a third of the data as a test set, you just need a large enough test set to measure performance.

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+
+    label_columns = ["label"]
+
+    for col in categorical_features:
+        data[col] = data[col].astype("category")
+
+    data = balance_motifs(data, params)
+    data["label"][data["label"] == "TP"] = 1
+    data["label"][data["label"] == "FP"] = 0
+    data = data.astype({"label": "int16"})
```
kenissur: 1. Why int16 and not a boolean? 
2. This would do this in a single line, assuming you only have TP or FP (which you assume anyway)
data.loc[:, "label"] = data["label"] == "TP"

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+    features = categorical_features + non_categorical_features
+    if "is_mixed" in data.columns:
+        features.append("is_mixed")
+
+    label_columns = ["label"]
+
+    for col in categorical_features:
+        data[col] = data[col].astype("category")
+
+    data = balance_motifs(data, params)
```
kenissur: Shouldn't be called if params["balance_motifs"] is False

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+
+def get_precision_recall(
+    df, max_score, label, score="qual", cumulative=False, apply_log_trans=True
+):
+    precisions = []
+    recalls = []
+    fprs = []
+    for i in range(max_score):
+        if cumulative:
+            d_q = df[df[score].apply(np.floor) >= i]
```
kenissur: Why do you need the np.floor?

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+        d.to_parquet(savename)
+
+    with open(params["params_save_path"], "w", encoding="utf-8") as f:
+        json.dump(params, f)
+    
+    return
+
+
+def get_precision(df, label):
+    return (df["label"] == label).sum() / df.shape[0] if df.shape[0] > 0 else np.nan
```
kenissur:
```suggestion
    return (df["label"] == label).mean()
```
```
```

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+):
+    precisions = []
+    recalls = []
+    fprs = []
+    for i in range(max_score):
+        if cumulative:
+            d_q = df[df[score].apply(np.floor) >= i]
+        else:
+            d_q = df[df[score].apply(np.floor) == i]
+        precisions.append(get_precision(d_q, label))
```
kenissur: looks like you don't need the precisions list, just this value

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+        recalls.append(get_recall(d_q, df, label))
+        if precisions[-1] == np.nan:
+            fprs.append(np.none)
+        elif apply_log_trans:
+            qual = -10 * np.log10(1 - precisions[-1])
+            qual = min(qual, max_score)
+            fprs.append(qual)
+        else:
+            fprs.append(precisions[-1])
+    return fprs, recalls
```
kenissur: I don't like the fact that the scores are implicit, I prefer to create a dataframe with theses values and the score thresholds in one place

ugvc/mrd/bqsr_train_utils.py
```diff
@@ -0,0 +1,807 @@
...
+    reference_fasta: str = None,
+    flow_order: str = DEFAULT_FLOW_ORDER,
+    training_chrom: str = "chr20",
+):
+    # init
+    os.makedirs(out_path, exist_ok=True)
+    if len(out_basename) > 0 and not out_basename.endswith("."):
+        out_basename += "."
+
+    params = {}
```
kenissur: I think this params dictionary structure is bad design. It does not allow for proper interpretation of the code by the IDE, and it contains different types of arguments. You have a combination of paths to save to, input files, and parameters for the model.

I suggest the following changes:
1. The input files like hom_snv_featuremap and single_substitution_featuremap should be their own parameters
2. Arguments to the XGBoost model should be a "**xgboost_args" dictionary that you can then pass as-is to the model
xgb_classifier = xgb.XGBClassifier(**xgboost_args)
Right now you have a combination of input parameters and harcoded values. You can have hard coded defaults for convenience.
3. Consider refactoring this into a class like Doron does

ugvc/pipelines/mrd/bqsr_inference.py
```diff
@@ -0,0 +1,97 @@
+#!/env/python
+# Copyright 2022 Ultima Genomics Inc.
```
kenissur: 2023

ugvc/pipelines/mrd/bqsr_inference.py
```diff
@@ -0,0 +1,97 @@
...
+        default="",
+        required=False,
+        help="""basename of output files""",
+    )
+    parser.add_argument(
+        "-r",
+        "--reference_fasta",
+        type=str,
+        help="""reference fasta, only required for motif annotation most likely
+         gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.fasta but it must be localized""",
```
kenissur: I'd remove the second line

ugvc/pipelines/mrd/bqsr_train.py
```diff
@@ -0,0 +1,100 @@
...
+        default="",
+        required=False,
+        help="""basename of output files""",
+    )
+    parser.add_argument(
+        "-r",
+        "--reference_fasta",
+        type=str,
+        help="""reference fasta, only required for motif annotation most likely
+         gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.fasta but it must be localized""",
```
kenissur: I'd remove this second line

ugvc/reports/bqsr_train_report.ipynb
```diff
@@ -0,0 +1,971 @@
...
+   "source": [
+    "# load model, data and params\n",
+    "xgb_classifier = joblib.load(model_file)\n",
+    "X_test = pd.read_parquet(X_file)\n",
+    "y_test = pd.read_parquet(y_file)\n",
+    "params = None\n",
+    "with open(params_file, 'r') as f:\n",
+    "    params = json.load(f)\n",
+    "\n",
+    "# for debug:\n",
```
kenissur: I'd remove this part, or at least remove the specific file names and paths

ugvc/reports/bqsr_train_report.ipynb
```diff
@@ -0,0 +1,971 @@
+{
```
kenissur: Please reset the notebook and recommit

kenissur (changes requested):
Hi Rina. There's quite a bit to work on, I think we'd best work on some refactoring together.

In addition to the inline comments, we need tests for these pipelines, and to make sure they run on both balanced ePCR featuremaps and on standard ones.

ugvc/mrd/bqsr_inference_utils.py
rinashalibo: move to inside main function

ugvc/mrd/bqsr_inference_utils.py
rinashalibo: make it a class

ugvc/mrd/bqsr_inference_utils.py
rinashalibo: chunk = []

ugvc/mrd/bqsr_inference_utils.py
rinashalibo: if we remove the ifs on the contigs = ...
need to make sure that the fetch from featuremap on these chrs is empty or process_chunk will not fail

ugvc/mrd/bqsr_train_utils.py
rinashalibo: decide training set size from all genome, ad decide the test set size

ugvc/mrd/bqsr_inference_utils.py
rinashalibo: will leave the code as is, it is more efficient to pre-allocate

ugvc/mrd/bqsr_inference_utils.py
rinashalibo: done

ugvc/mrd/bqsr_train_utils.py
rinashalibo: done

ugvc/mrd/bqsr_train_utils.py
rinashalibo: done

ugvc/pipelines/mrd/bqsr_inference.py
rinashalibo: done

ugvc/pipelines/mrd/bqsr_inference.py
rinashalibo: done

ugvc/pipelines/mrd/bqsr_train.py
rinashalibo: done

ugvc/reports/bqsr_train_report.ipynb
rinashalibo: done

ugvc/reports/bqsr_train_report.ipynb
rinashalibo: done

PR #49, opened 2023-07-25
pileup based read features

doron-st:


ugvc/__main__.py
```diff
@@ -12,23 +12,31 @@
     sys.path.insert(0, path)
 
 from ugvc.cnv import filter_sample_cnvs
-
+from ugvc.joint import compress_gvcf
```
doron-st, line 15: this is the precommit's doing. 
I don't know how to prevent it from changing stuff

ugvc/pipelines/mrd/pileup_based_read_features.py
```diff
@@ -0,0 +1,332 @@
...
+
+    FEATURE_NAMES = [f.name for f in FEATURES_LIST]
+    MAX_REGION_SIZE = 100000
+    PADDING = 300
+    AF_THRESHOLD = 0.2
+
+    LAST_POSITION = 0
+    LAST_CHR = ""
+
+    def __init__(self, args):
```
kenissur, line 86: Wouldn't you want an interface you can invoke from another python function instead of only as a standalone script?

kenissur (approved):
Looks great, just one minor comment/question.

ugvc/pipelines/mrd/pileup_based_read_features.py
doron-st: You can invoke it from another python function:
 PileupBasedReadFeatures(args).process()
 
Do you mean you would want to specify the arguments explicitly?

kenissur:
That’s what I meant, but I don’t insist at all

***@***.***<https://www.ultimagenomics.com/>
Itai Rusinek​​​​
Sr. Researcher
E:
***@***.***

ugvc/pipelines/mrd/pileup_based_read_features.py
doron-st: ok, done

PR #58, opened 2023-08-09
add a db_control option to intersect_featuremap_with_signature

GatKrieger:
needed for updating mrd_featuremap.wdl

### DIFF
--- a/ugvc/mrd/mrd_utils.py
+++ b/ugvc/mrd/mrd_utils.py
@@ -92,6 +92,28 @@ def _safe_tabix_index(vcf_file: str):
     return vcf_file
 
 
+def _validate_info(x):
+    """
+    Validate INFO field in a VCF record.
+    In some cases it is a string of boolean ("TRUE"/"FALSE")
+    In the case of long hmers it's numeric: if the hmer >= 7
+    it appears in the INFO field as a tuple of (hmer, hmer_length),
+    then the function should return True
+    """
+    MAX_HMER_LENGTH = 7
+    if isinstance(x, bool):
+        return x
+    if isinstance(x, str):
+        if x.lower() == "true":
+            return True
+        if x.lower() == "false":
+            return False
+        if x.isnumeric():
+            return bool(float(x) >= MAX_HMER_LENGTH)
+        raise ValueError(f"Cannot convert {x} to bool")
+    return None
+
+
 def read_signature(  # pylint: disable=too-many-branches
     signature_vcf_files: list[str],
     output_parquet: str = None,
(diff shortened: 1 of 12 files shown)


GatKrieger:
I need help with changing the --is_matched flag in other functions: read_signature, featuremap_to_dataframe

kenissur (approved):

ugvc/mrd/mrd_utils.py
```diff
@@ -92,6 +92,16 @@ def _safe_tabix_index(vcf_file: str):
     return vcf_file
 
 
+def _validate_info(x):
```
kenissur, line 95: First check if the input is boolean and if so return it, then checks whether it's a string before doing string operations, and please add a docstring and explain what this is for.

ugvc/mrd/mrd_utils.py
```diff
@@ -209,6 +219,7 @@ def read_signature(  # pylint: disable=too-many-branches
             logger.debug(f"Reading x_columns: {x_columns}")
 
             for j, rec in enumerate(variant_file):
+                # print(rec)
```
kenissur: remove

ugvc/mrd/mrd_utils.py
```diff
@@ -235,7 +246,8 @@ def read_signature(  # pylint: disable=too-many-branches
                             else np.nan,
                             rec.info["SOR"] if "SOR" in rec.info else np.nan,
                         ]
-                        + [c in rec.info for c in genomic_region_annotations]
+                        # [c in rec.info for c in genomic_region_annotations]
```
kenissur: remove commented line

ugvc/mrd/mrd_utils.py
```diff
@@ -301,7 +313,7 @@ def read_signature(  # pylint: disable=too-many-branches
                 for variant_file in f_bw:
                     variant_file.close()
         df_sig = pd.concat(df_list).astype({"coverage": int})
-
+    df_sig["hmer"] = np.nan  # it read TRUE/FALSE in some cases, so we reset it to nan
```
kenissur: 1. Please explain
2. I don't like this patch, if the issue is that sometime you get string values then find a way to force-convert to float (as in, convert to float if possible and it not set to NaN), pretty sure pandas can do that

ugvc/mrd/mrd_utils.py
```diff
@@ -808,22 +787,28 @@ def prepare_data_from_mrd_pipeline(
```
kenissur, line 826: Please make sure at least one of db_control_signatures_vcf_files, control_signatures_vcf_files, matched_control_signatures_vcf_files is not None
Recommend adding a test that gives just one of these to make sure it doesn't fail

ugvc/mrd/mrd_utils.py
```diff
@@ -93,13 +93,24 @@ def _safe_tabix_index(vcf_file: str):
...
+    """
+    if isinstance(x, bool):
+        return x
+    if isinstance(x, str):
+        if x.lower() == "true":
+            return True
+        if x.lower() == "false":
+            return False
+        if x.isnumeric():
+            return True
```
kenissur: if x==0 then I would expect False, so maybe "return bool(float(x))"?

PR #60, opened 2023-08-10
added illustration png that was ommited because of gitignore

kenissur:
The illustration file was absent from the PR, which causes the report generation to fail. I built a docker from this version and it works.


GatKrieger:
Please edit it so it matches the current loop sequence and LIG/HYB terminology

GatKrieger (approved):

PR #64, opened 2023-08-14
Now the GH action fails even if the intermediate command fails

ilyasoifer:
We discovered that the CI action does not fail when unit tests are failing and this caused some accumulation of failed unit tests in the master. This PR fixes that together with two PRs that fix the broken tests


ilyasoifer:
I am assuming only my commits require review here, but @kenissur  and @GatKrieger  - please let me know if otherwise

kenissur (approved):

kenissur:
That's true, they were reviewed separately already.

I don't understand where you made a change but I see it works :)

test/unit/reports/test_methyldackel_create_qc_report.py
```diff
@@ -8,33 +7,19 @@ def test_methyldackel_create_qc_report(tmpdir):
...
-        "--template",
-        "classic",
-        "--no-input",
-        "--output",
-        "methyldackel_qc_report.html",
-    ]
-    assert subprocess.check_call(cmd, cwd=tmpdir) == 0
\ No newline at end of file
+    cmd2 = (
+        f"jupyter nbconvert --to html {papermill_out} " f"--template classic --no-input --output {base_file_name}.html"
```
mayalevy: You can make it one string, now it is 2 concatenated strings

.github/workflows/python-package-conda.yml
```diff
@@ -20,32 +17,32 @@ jobs:
...
       - name: Install dependencies
         run: |
-          pip install flake8
-          conda env create --file setup/other_envs/ucsc.yml --name ucsc
-          conda env create -f setup/other_envs/cutadaptenv.yml --name cutadaptenv
-          conda env create --file setup/other_envs/cnmops.yml --name cn.mops
-          conda env create --file setup/environment.yml --name genomics.py3
+          pip install flake8 
+          conda env create --file setup/other_envs/ucsc.yml --name ucsc 
+          conda env create -f setup/other_envs/cutadaptenv.yml --name cutadaptenv
```
mayalevy, line 29: Maybe remove all these spaces?

mayalevy (approved):

.github/workflows/python-package-conda.yml
mayalevy: Can you explain how you resolved that? 
I guess by removing these line
   defaults:
      run:
        shell: bash -l {0}
        
but why?

.github/workflows/python-package-conda.yml
```diff
@@ -20,32 +17,32 @@ jobs:
           conda-channels: conda-forge, bioconda, r, defaults
       - name: Setup file system
         run: |
-          sudo mkdir -p /data
+          sudo mkdir -p /data && \
```
doron-st, line 20: Do we still need the &&?

.github/workflows/python-package-conda.yml
doron-st: I read something about github using set -e by default, but not when using a non-default shell, which isprobably what we did here.

doron-st (commented):
Tiny comment

PR #66, opened 2023-08-14
add cnmops env

TammushB:


test/system/test_create_reads_count_cohort_matrix.py
```diff
@@ -29,7 +29,8 @@ def test_create_reads_count_cohort_matrix(tmpdir):
         "--vanilla",
         script_path,
         "-samples_read_count_files_list",
-        rc_files_list
+        rc_files_list,
+        "--save_csv"
     ]
     assert subprocess.check_call(cmd, cwd=tmpdir) == 0
     assert filecmp.cmp(out_file, expected_out_file)
```
ilyasoifer: Better np.allclose like in test_coverage_analysis. Float's can sometimes change with no apparent reason

ilyasoifer (commented):
Just a minor comment regarding the tests

test/system/test_create_reads_count_cohort_matrix.py
TammushB: @ilyasoifer to implement with np.allclose we need to compare only numerical columns. 
actually, the only test that outputs float and not integer is  _test_normalize_reads_count.py_ .
anyway, I implemented this also for all other tests. I guess that in the future we will find ourselves comparing int to float ;)

ilyasoifer (approved):

PR #68, opened 2023-08-15
[BIOIN-1209] Added option for coverage collection to only collect depth files

ilyasoifer:


.gitattributes
```diff
@@ -12,6 +12,7 @@
 *.bw filter=lfs diff=lfs merge=lfs -text
 *.interval_list filter=lfs diff=lfs merge=lfs -text
 *.bedGraph filter=lfs diff=lfs merge=lfs -text
+*.bedgraph filter=lfs diff=lfs merge=lfs -text
```
kenissur, line 15: maybe this line was duplicates by mistake?

ugvc/pipelines/coverage_analysis.py
```diff
@@ -241,6 +341,7 @@ def run_full_coverage_analysis(
         "-a",
         "-J",
         ref_str,
+        ref_str,
```
kenissur: duplicated by mistake again?

test/system/test_coverage_analysis.py
```diff
@@ -31,3 +32,26 @@ def test_coverage_analysis(tmpdir):
     df = pd.read_hdf(pjoin(tmpdir, "170201-BC23.coverage_stats.q0.Q0.l0.h5"), "percentiles")
     df_ref = pd.read_parquet(f_ref)
     assert np.allclose(df.fillna(-1), df_ref.fillna(-1))
+
+
+def test_coverage_collection(tmpdir):
```
kenissur, line 37: I didn't know this tmpdir input option - cool!

doron-st (approved):
Can you add a few words on the difference between the full analysis and the limited analysis?
Not sure I got it

kenissur (approved):
Looks good, I should have separated it this way originally :)

Just comments about possible line duplications

.gitattributes
ilyasoifer: bed**g**raph versus bed**G**raph.

test/system/test_coverage_analysis.py
ilyasoifer: :+1

ugvc/pipelines/coverage_analysis.py
ilyasoifer: Fixed

PR #70, opened 2023-08-27
Maya bioin 1211 fix run no gt snv only

mayalevy:
gsutil cp gs://cromwell-backend-ultima-data-307918/cromwell-execution/VariantCallingEvaluation/b152dbdd-98f9-4790-98da-e7f442b63ee5/call-CreateNoGTReport/infer_hg007_model_5_60_v_v_no_gt_report.html .

This is how it looks when there are snps only. Maybe better to remove that at all? Please take a look. I think I like that but maybe it is confusing?

### DIFF
--- a/ugvc/pipelines/run_no_gt_report.py
+++ b/ugvc/pipelines/run_no_gt_report.py
@@ -313,58 +313,107 @@ def run_somatic_analysis(arg_values):
 
     def _get_signatures_from_sig_log(log_file):
         line = log_file.readline()
-        while "Composition After Add-Remove" not in line:
+        while line and "Composition After Add-Remove" not in line:
             line = log_file.readline()
 
-        signatures = log_file.readline().strip("\n").split()
-        activity_values = log_file.readline().strip("\n").split()[1:]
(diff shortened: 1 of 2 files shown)


ugvc/pipelines/run_no_gt_report.py
```diff
@@ -313,58 +313,107 @@ def run_somatic_analysis(arg_values):
...
+        Analyze.cosmic_fit(samples=vcf_dirname,
+                           output=os.path.join(output_dir,'sbs'),
+                           input_type="vcf",
+                           context_type="96",
+                           genome_build=arg_values.reference_name,
+                           cosmic_version=arg_values.cosmic_version)
+    except ValueError as e:
+        # check if the exception was caused because there are no snp variants at all
+        variant_count = _check_number_of_variants(arg_values.input_file, "snp")
+        if variant_count > 0:
```
ilyasoifer, line 357: Maybe print the number of variants anyway in this case? Will help debugging

ilyasoifer (approved):

PR #74, opened 2023-09-04
Rinas/add features ml as input

rinashalibo:


kenissur (changes requested):
Hi Rina,

There is a way to give multiple arguments in argparse, I think something like "nargs='+'", then you indeed get a list as input. Please do that and revert the changes in ugvc/mrd/bqsr_train_utils.py

PR #81, opened 2023-09-13
Srsnv_motif_balancing_feature

kenissur:
Fixed motif balancing feature + minor fixes


ugvc/mrd/srsnv_training_utils.py
```diff
@@ -165,10 +165,14 @@ def prepare_featuremap_for_model(
...
     if featuremap_entry_number < total_size:
         logger.warning(
-            f"featuremap_entry_number={featuremap_entry_number} > training_set_size={train_set_size}"
+            "Requested training and test set size cannot be met - insufficient data"
+            f"featuremap_entry_number={featuremap_entry_number} < training_set_size={train_set_size}"
             f"+ test_set_size={test_set_size if test_set_size else 0}"
-            "\nbehavior is undefined",
-        )  # TODO in this case we should make sure we still have a test set (and behavior should be defined)
+        )
+        if test_set_size is not None:
```
doron-st: why would it be None?

ugvc/mrd/srsnv_training_utils.py
```diff
@@ -165,10 +165,14 @@ def prepare_featuremap_for_model(
...
+            train_set_size = np.floor(featuremap_entry_number * (train_set_size / total_size))
```
doron-st, line 173: the total_size is the total size of the featuremap?
So if test-size is None we use all remaining rows as test?

test/unit/mrd/test_srsnv_train_utils.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def test_prepare_featuremap_for_model_with_motif_balancing(tmpdir):
+    """Test that downsampling training-set works as expected"""
+
+    input_featuremap_vcf = pjoin(
+        inputs_dir,
+        "333_LuNgs_08.annotated_featuremap.vcf.gz",
+    )
+    balanced_sampling_info_fields = ["trinuc_context", "is_forward"]
```
doron-st, line 50: do we still have trinuc_contextas a feature? I'm confused.

test/unit/mrd/test_srsnv_train_utils.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+        assert __count_variants(downsampled_training_featuremap_vcf) == train_set_size
+
+        balanced_sampling_info_fields_counter = defaultdict(int)
+        with pysam.VariantFile(downsampled_training_featuremap_vcf) as fmap:
+            for record in fmap.fetch():
+                balanced_sampling_info_fields_counter[
+                    tuple(record.info.get(info_field) for info_field in balanced_sampling_info_fields)
+                ] += 1
+        assert sum(balanced_sampling_info_fields_counter.values()) == train_set_size
```
doron-st, line 69: Let's add an assertion about the uniformity

ugvc/pipelines/mrd/srsnv_training.py
```diff
@@ -83,6 +83,12 @@ def parse_args(argv: list[str]) -> argparse.Namespace:
         type=str,
         help="""comma separated list of categorical features for ML classifier """,
     )
+    parser.add_argument(
+        "--balanced_sampling_info_fields",
+        type=str,
+        help="comma separated list of categorical features to be used for balanced sampling of the TP training set"
+        " to eliminate prior distribution bias (e.g. 'trinuc_context_with_alt,is_forward')",
```
doron-st, line 90: remind me, the context of forward and reverse reads on the same snv is the same or revcom?

doron-st (commented):
Looks mostly good. A few questions and a small request on the test

doron-st:
A general question about balancing.
Do you balance only the TPs or all?

test/unit/mrd/test_srsnv_train_utils.py
kenissur: Oh. There's an older file... do you insist I change it?

ugvc/mrd/srsnv_training_utils.py
kenissur: No reason, fixing

ugvc/mrd/srsnv_training_utils.py
kenissur: total_size = train_set_size + test_set_size

Removing the option to have None

ugvc/pipelines/mrd/srsnv_training.py
kenissur: The same, that's why I added is_forward

test/unit/mrd/test_srsnv_train_utils.py
kenissur: How is that? :)

ugvc/mrd/srsnv_training_utils.py
doron-st: Wait. 
if featuremap_entry_number is the size of the featuremap (in the region)
and total_size is train+test 
this calculation seems weird isn't it?

kenissur: I think it's okay. I am handling the case where we asked for more training data than the total number of available entries, so in this case I am using approximately all the data, but maintaining the ratio of train_set_size to test_set_size

ugvc/mrd/srsnv_training_utils.py
```diff
@@ -185,28 +189,23 @@ def prepare_featuremap_for_model(
...
         np.random.seed(random_seed)
     if balanced_sampling_info_fields:
-        # TODO test this code and add a test, I wrote an implementation but didn't even get to run it
-        logger.warning(
-            "Data motif balancing not tested, use at your own risk"
-            f" (balanced_sampling_info_fields={balanced_sampling_info_fields})"
-        )
         # count the number of entries with each combination of info fields to balance by
         balanced_sampling_info_fields_counter = defaultdict(int)
         with pysam.VariantFile(intersect_featuremap_vcf) as fmap:
```
doron-st: Wouldn't it be better to count while in the previous general counting loop? for efficiency reasons.

ugvc/mrd/srsnv_training_utils.py
```diff
@@ -185,28 +189,23 @@ def prepare_featuremap_for_model(
...
             for record in fmap.fetch():
                 balanced_sampling_info_fields_counter[
-                    (record.info.get(info_field) for info_field in balanced_sampling_info_fields_counter)
+                    tuple(record.info.get(info_field) for info_field in balanced_sampling_info_fields)
                 ] += 1
         # determine sampling rate per group (defined by a combination of info fields)
         total_size_per_group = total_size // len(balanced_sampling_info_fields_counter)
```
doron-st: I'm confused by this calculation.
Isn't it more natural to count the fraction of events with this info-fields from the total?
And then use the inverse of that to drive the balancing?

ugvc/mrd/srsnv_training_utils.py
```diff
@@ -185,28 +189,23 @@ def prepare_featuremap_for_model(
...
-            k: total_size_per_group / v * overhead_factor for k, v in balanced_sampling_info_fields_counter.items()
+            k: (total_size_per_group / v) * overhead_factor for k, v in balanced_sampling_info_fields_counter.items()
         }
         for k, v in downsampling_rate.items():
             logger.debug(f"downsampling_rate for {k} = {v}")
-            if v > 1:
+            if v > overhead_factor:
                 logger.warning(f"downsampling_rate for {k} = {v} > 1, result will contain less data than expected")
-                downsampling_rate[k] = 1
+                downsampling_rate[k] = min(1, downsampling_rate[k])
```
doron-st: Still not sure what's going on here. sorry...
If we want to make sure ds rate is not larger than 1, why don't we do this explicitly?
downsampling_rate[k] = min(1, downsampling_rate[k])

ugvc/mrd/srsnv_training_utils.py
kenissur: You're right, fixed.

ugvc/mrd/srsnv_training_utils.py
kenissur: 1. I removed the overhead factor for simplicity
2. total_size_per_group is the total number of the entries we want in the output, and I am dividing it by the total number of groups we are balancing across to get total_size_per_group
3. For each group I am then dividing total_size_per_group by the number of total number of featuremap entries in the group to get the rate at which I want to downsample. I made the variable names more explicit.

ugvc/mrd/srsnv_training_utils.py
kenissur: It is done explicitly in the next line

ugvc/pipelines/mrd/srsnv_training.py
doron-st: So your intent is that in the end we will have the same number of +/- reads per motif?
I'm not sure why wouldn't it occur naturally in germline variants.

doron-st (approved):
Excellent work

PR #86, opened 2023-09-21
Created a create_hom_snv_featuremap python pipeline, it does the same as the previous wdl task except the minimum coverage requirement that is now dynamic (the lower between the requested coverage and the median coverage)

kenissur:


ugvc/pipelines/mrd/create_hom_snv_featuremap.py
```diff
@@ -0,0 +1,81 @@
...
+        "--requested_min_coverage",
+        type=int,
+        default=20,
+        help="Minimum coverage requested for locus to be propagated to the output. If the median coverage "
+        "is lower than this value, the median coverage will be used as the minimum coverage instead.",
+    )
+    parser.add_argument(
+        "--min_af",
+        type=float,
+        default=0.7,
```
doron-st, line 59: Not sure 0.7 is high enough.
loci with 70/30 distribution sound problematic, and not great examples for good reads.

doron-st (approved):
Very nice. I only questioned the 70% threshold.

PR #90, opened 2023-09-28
Rinas/srsnv fix categorial encoding

rinashalibo:


doron-st (changes requested):
Still need to get the X_train_path from srsnv_inference argparse.

PR #91, opened 2023-09-28
fix categorical encoding and save X_train path

rinashalibo:
fix loading categories

add arg X_train to srsnv pipeline


doron-st (approved):
LGTM

kenissur:
Rebased an pushed (was missing a new commit fixing an unrelated bug I caused)

PR #92, opened 2023-10-02
Gat/ssqc report fixes

GatKrieger:
After Omry's comments


ugvc/scripts/sorter_to_h5.py
```diff
@@ -119,6 +136,18 @@ def sorter_to_h5(
     ]
     h5_dict["RawWgsMetrics"] = {k: h5_dict["RawWgsMetrics"][k] for k in key_order}
 
+    # Q20_PF_BASES and Q30_PF_BASES
+    h5_dict["QualityYieldMetricsFlow"]["Q20_BASES"] = round(
```
doron-st, line 140: I think it's concerning to rename he percentage of pass-filter bases as just bases.
Do we have another choice?

doron-st (commented):
just one comment

ugvc/scripts/sorter_to_h5.py
GatKrieger: That's what we should do according to Omer.

doron-st (approved):
LGTM

PR #103, opened 2023-10-31
collecting data from base_coverage is optional

GatKrieger:
for the terra PR: gat/add_ssqc_to_trim_align_sort

### DIFF
--- a/ugvc/scripts/sorter_to_h5.py
+++ b/ugvc/scripts/sorter_to_h5.py
@@ -63,7 +63,7 @@ def sorter_to_h5(
             h5_dict[row["H5_key"]][row["H5_item"]] = json_data[row["key"]]
 
     # extract coverage stats
-    if not json_data["base_coverage"] == {}:  # if base_coverage is not empty
+    if "base_coverage" in json_data and len(json_data["base_coverage"]) > 0:  # if base_coverage is not empty
         df_coverage_histogram = (
             pd.concat(
                 (
--- a/ugvc/scripts/sorter_to_h5.py
+++ b/ugvc/scripts/sorter_to_h5.py
@@ -63,51 +63,56 @@ def sorter_to_h5(
             h5_dict[row["H5_key"]][row["H5_item"]] = json_data[row["key"]]
 
     # extract coverage stats
-    df_coverage_histogram = (
-        pd.concat(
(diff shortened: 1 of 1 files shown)


ugvc/scripts/sorter_to_h5.py
```diff
@@ -63,51 +63,56 @@ def sorter_to_h5(
             h5_dict[row["H5_key"]][row["H5_item"]] = json_data[row["key"]]
 
     # extract coverage stats
-    df_coverage_histogram = (
-        pd.concat(
-            (pd.DataFrame(json_data["base_coverage"][key], columns=[key]) for key in json_data["base_coverage"].keys()),
-            axis=1,
+    if not json_data["base_coverage"] == {}:  # if base_coverage is not empty
```
kenissur: To be on the safe side I would do
`if "base_coverage" in json_data and len(json_data["base_coverage"]) == 0`

I wouldn't rely on the fact that Ohad creates an empty field

ugvc/scripts/sorter_to_h5.py
```diff
@@ -63,51 +63,56 @@ def sorter_to_h5(
...
+        # add the F80, F90, F95 metrics to RawWgsMetrics
+        h5_dict["RawWgsMetrics"]["FOLD_80_BASE_PENALTY"] = (
+            df_percentiles.loc["Q50", "Genome"] / df_percentiles.loc["Q20", "Genome"]
+        )
+        h5_dict["RawWgsMetrics"]["FOLD_90_BASE_PENALTY"] = (
+            df_percentiles.loc["Q50", "Genome"] / df_percentiles.loc["Q10", "Genome"]
+        )
+        h5_dict["RawWgsMetrics"]["FOLD_95_BASE_PENALTY"] = (
+            df_percentiles.loc["Q50", "Genome"] / df_percentiles.loc["Q5", "Genome"]
+        )
```
kenissur, line 115: I assume none of this has changed, right? It looks weird on GitHub

kenissur (approved):
Just one minor suggestion

ugvc/scripts/sorter_to_h5.py
GatKrieger: this part (add the F80, F90, F95 metrics to RawWgsMetrics) is part of the base_coverage condition. if base_coverage is empty, it takes these metrics from the sorter.csv

PR #111, opened 2023-11-26
SRSNV - individual bases instead of motif encoding

rinashalibo:


rinashalibo:
closing for rebase with master

PR #112, opened 2023-11-26
SRSNV - individual bases instead of motif encoding

rinashalibo:


ugvc/mrd/featuremap_utils.py
```diff
@@ -43,6 +43,12 @@ class FeatureMapFields(Enum):
     X_CIGAR = "X_CIGAR"
     PREV_3bp = "prev_3bp"
     NEXT_3bp = "next_3bp"
+    PREV_1 = "next_1sbp"
```
doron-st: what is sbp? 
why not simply 'next_1'?

ugvc/mrd/featuremap_utils.py
```diff
@@ -43,6 +43,12 @@ class FeatureMapFields(Enum):
...
+    PREV_2 = "next_2sbp"
+    PREV_3 = "next_3sbp"
+    NEXT_1 = "prev_1sbp"
```
doron-st: NEXT_1 should be "next_1" not "prev_1"

ugvc/mrd/featuremap_utils.py
```diff
@@ -254,6 +272,24 @@ def edit_vcf_header(self, header: pysam.VariantHeader) -> pysam.VariantHeader:
             f'Number=1,Type=String,Description="{self.motif_length_to_annotate}'
             'bases in the reference after variant">'
         )
+        header.add_line(
+            f"##INFO=<ID={self.NEXT_1}," f'Number=1,Type=String,Description="1 base in the reference after variant">'
+        )
+        header.add_line(
+            f"##INFO=<ID={self.NEXT_2}," f'Number=1,Type=String,Description="2 bases in the reference after variant">'
```
doron-st: NEXT_2 is the second base right? here it say '2 bases'

ugvc/mrd/featuremap_utils.py
```diff
@@ -305,6 +341,12 @@ def process_records(self, records: list[pysam.VariantRecord]) -> list[pysam.Vari
                 record.info[self.NEXT_N_BP] = ref_around_snv[
                     central_base_ind + 1 : central_base_ind + self.motif_length_to_annotate + 1
                 ]
+                record.info[self.NEXT_1] = ref_around_snv[central_base_ind + 1]
```
doron-st: I think this is better be done in a loop.
can save the array of 6 feature names.

ugvc/mrd/srsnv_inference_utils.py
```diff
@@ -126,7 +126,13 @@ def process_records(self, records: list[pysam.VariantRecord]) -> list[pysam.Vari
         """
         # Convert list of variant records to a pandas DataFrame
         columns = self.numerical_features + self.categorical_features
-        data = [{column: variant.info[column] for column in columns} for variant in records]
+        data = [
```
doron-st: I understand what you are trying to do here.
But I wonder if there is a simpler syntax.
Can we read the info columns like before, and then add ref and alt with simple apply/append?

ugvc/mrd/featuremap_utils.py
```diff
@@ -211,12 +217,24 @@ def __init__(
         self.CYCLE_SKIP_FLAG = FeatureMapFields.IS_CYCLE_SKIP.value
         self.PREV_N_BP = f"prev_{self.motif_length_to_annotate}bp"
         self.NEXT_N_BP = f"next_{self.motif_length_to_annotate}bp"
+        self.NEXT_1 = "next_1sbp"
```
doron-st: This can also be reduced to a loop I think

ugvc/mrd/featuremap_utils.py
```diff
@@ -211,12 +217,24 @@ def __init__(
...
+        self.PREV_1 = "prev_1sbp"
+        self.PREV_2 = "prev_2sbp"
+        self.PREV_3 = "prev_3sbp"
         self.info_fields_to_add = [
             self.TRINUC_CONTEXT_WITH_ALT,
             self.HMER_CONTEXT_REF,
             self.HMER_CONTEXT_ALT,
             self.PREV_N_BP,
             self.NEXT_N_BP,
+            self.NEXT_1,
```
doron-st: if we have a list of features, we could just join the lists here:
info_fields_to_add = [F1, F2, ... ] + ref_bases_feature_names

doron-st (changes requested):
Looks correct.
Please see comments, mainly about coding style.

ugvc/mrd/featuremap_utils.py
rinashalibo: s for single, as opposed to next_3

ugvc/mrd/featuremap_utils.py
rinashalibo: wow, sorry - my mistake, happy you noticed, fixed

ugvc/mrd/featuremap_utils.py
rinashalibo: corrected

ugvc/mrd/srsnv_inference_utils.py
rinashalibo: tried it, and seems this is the best way to create a dictionary from two dictionaries

ugvc/mrd/featuremap_utils.py
```diff
@@ -211,12 +217,24 @@ def __init__(
...
         self.PREV_N_BP = f"prev_{self.motif_length_to_annotate}bp"
         self.NEXT_N_BP = f"next_{self.motif_length_to_annotate}bp"
+        self.NEXT_1 = "next_1sbp"
+        self.NEXT_2 = "next_2sbp"
+        self.NEXT_3 = "next_3sbp"
+        self.PREV_1 = "prev_1sbp"
+        self.PREV_2 = "prev_2sbp"
+        self.PREV_3 = "prev_3sbp"
         self.info_fields_to_add = [
             self.TRINUC_CONTEXT_WITH_ALT,
```
kenissur, line 222: Let's remove TRINUC_CONTEXT_WITH_ALT now that it's redundant

ugvc/mrd/featuremap_utils.py
```diff
@@ -254,6 +272,24 @@ def edit_vcf_header(self, header: pysam.VariantHeader) -> pysam.VariantHeader:
             f'Number=1,Type=String,Description="{self.motif_length_to_annotate}'
             'bases in the reference after variant">'
         )
+        header.add_line(
+            f"##INFO=<ID={self.NEXT_1}," f'Number=1,Type=String,Description="1 base in the reference after variant">'
+        )
```
kenissur: @doron-st taught me that it's better to use .add_meta (you can find his examples in the repo) than .add_line

ugvc/mrd/featuremap_utils.py
```diff
@@ -211,12 +217,24 @@ def __init__(
         self.CYCLE_SKIP_FLAG = FeatureMapFields.IS_CYCLE_SKIP.value
         self.PREV_N_BP = f"prev_{self.motif_length_to_annotate}bp"
         self.NEXT_N_BP = f"next_{self.motif_length_to_annotate}bp"
```
kenissur: Please remove self.NEXT_N_BP and self.PREV_N_BP

ugvc/mrd/featuremap_utils.py
kenissur: Agree, please use the input arg motif_length_to_annotate instead of hard coding 3

ugvc/mrd/featuremap_utils.py
```diff
@@ -254,6 +272,24 @@ def edit_vcf_header(self, header: pysam.VariantHeader) -> pysam.VariantHeader:
...
+        )
+        header.add_line(
+            f"##INFO=<ID={self.PREV_1}," f'Number=1,Type=String,Description="1 base in the reference before variant">'
+        )
+        header.add_line(
+            f"##INFO=<ID={self.PREV_2}," f'Number=1,Type=String,Description="2 bases in the reference before variant">'
+        )
+        header.add_line(
+            f"##INFO=<ID={self.PREV_3}," f'Number=1,Type=String,Description="3 bases in the reference before variant">'
+        )
```
kenissur, line 258: Should be a loop as well

ugvc/mrd/srsnv_inference_utils.py
kenissur: Suggestion, untested:
`
from ugvc.dna.format import ALT, CHROM, FILTER, POS, QUAL, REF
vcf_standard_column = [ALT, CHROM, FILTER, POS, QUAL, REF]
def _vcf_getter(variant, field):
    if field==ALT:
        return variant.alts[0]
    elif field==REF:
        return variant.ref
    elif field==CHROM:
        return variant.chrom
    elif field==POS:
        return variant.pos
    else:
        return variant.info.get(field, None)

data = [
{column: _vcf_getter(variant, column) for column in columns}}
            for variant in records
]`

In any case please use these constants

kenissur (changes requested):

ugvc/mrd/featuremap_utils.py
rinashalibo: its still used for balancing, I think it would be useful to keep it, WDYT?

ugvc/mrd/featuremap_utils.py
rinashalibo: done

ugvc/mrd/featuremap_utils.py
rinashalibo: done

ugvc/mrd/srsnv_inference_utils.py
rinashalibo: done

ugvc/mrd/featuremap_utils.py
rinashalibo: done

ugvc/mrd/featuremap_utils.py
rinashalibo: done

ugvc/mrd/featuremap_utils.py
rinashalibo: done

ugvc/mrd/featuremap_utils.py
kenissur: It needs to be replaced with prev_1,ref,next_1,alt in balanced sampling

kenissur (approved):
Approving, please just follow through on removing TRINUC_CONTEXT_WITH_ALT.

Best if @doron-st reviews too because I am not too sharp at this hour :)

ugvc/mrd/featuremap_utils.py
doron-st: It's not very harmfull to leave it for balancing I think.
Could also do it later on

doron-st (approved):
LGTM

PR #125, opened 2024-01-16
fix #pos treated as header in report description

doron-st:
image


ilyasoifer (approved):
# neg does not bother you?

doron-st:
> # neg does not bother you?

It didn't because it doesn't have '#'
But I agree that the missing '#' can be considered a mistake.
Will add it.

PR #128, opened 2024-02-02
Update Dockerfile.jbvc

ilyasoifer:
Added conda clean to make docker image smaller


ilyasoifer:
@TammushB and @mayalevy  - any chance you could review this? The goal is just to make the docker image slightly smaller

mayalevy (approved):
Looks good
Say, there is pip install mpld3==0.5.1
in line 30, can't we move it to all the other pip installs in the bottom of the env file?
It is not exactly its place here..

PR #131, opened 2024-02-06
[BIOIN-936] Working on python code for the updated training of filtering models

ilyasoifer:
This is a relatively large piece of work that is aimed to update our ability to train filtering models on the VCFs. 
It is inspired mostly by the work we did on filtering the FFPE VCFs, using FF VCFs as the ground truth. 
This PR is only about the creation of the labeled train data, with the main non-trivial task around the treatment of multiallelics and spanning deletions

Of all the changes I would appreciate review on filtering/training_prep.py (that prepares the labeled data) and spandel.py / multiallelics.py that represent multiallelic variants in specific way. 

Other changes are less important/cosmetic.


ugvc/filtering/multiallelics.py
```diff
@@ -0,0 +1,526 @@
...
+    cluster = []
+    for i in range(df.shape[0]):
+        if len(cluster) == 0 and del_length[i] == 0:
+            continue
+        if df.iloc[i]["pos"] > current_span:
+            if len(cluster) > 1:
+                for c in cluster:
+                    if c in multiallelic_locations:
+                        multiallelic_locations.remove(c)
+                results.append(cluster[:])
```
doron-st, line 47: is cluster[:] != cluster?

ugvc/filtering/multiallelics.py
```diff
@@ -0,0 +1,526 @@
...
+            cluster = []
+        if len(cluster) == 0:
+            cluster.append(i)
+        elif len(cluster) > 0 and SPAN_DEL in df.iloc[i]["alleles"]:
```
doron-st, line 53: Is SPAN_DEL == '*'?
Why do we need to check for it?
What happens if there is no such allele? we just ignore this variant?

ugvc/filtering/multiallelics.py
```diff
@@ -0,0 +1,526 @@
...
+            cluster.append(i)
+        current_span = max(current_span, del_length[i] + df.iloc[i]["pos"])
+    for m in multiallelic_locations:
+        results.append([m])
+    sorted_results = sorted(results)
```
doron-st, line 60: what are we sorting by here?

ugvc/filtering/multiallelics.py
```diff
@@ -0,0 +1,526 @@
...
+    pd.DataFrame
+        A dataframe with the same columns as the training set dataframe
+    """
+    record_to_nbr_dict = vcftools.header_record_number(call_vcf_header)
+    ignore_columns = ["gt_vcfeval", "alleles_vcfeval", "label", "sync"]
+
+    for ic in ignore_columns:
+        record_to_nbr_dict[ic] = 1
+
+    alleles = multiallelic_variant["alleles"]
```
doron-st, line 99: Does this include the reference allele?

ugvc/filtering/multiallelics.py
```diff
@@ -0,0 +1,526 @@
...
+        A row from the training set dataframe
+    call_vcf_header :
+        Header of the VCF : pysam.VariantHeader | pysam.VariantFile | str
+    ref : pyfaidx.Fasta
+        Reference
+
+    Returns
+    -------
+    pd.DataFrame
+        A dataframe with the same columns as the training set dataframe
```
doron-st: It's not clear what is the training-set dataframe refered to in here.

ugvc/filtering/multiallelics.py
```diff
@@ -0,0 +1,526 @@
...
+    # very rare case when there is a spanning deletion without an actual deletion (I guess GATK bug)
+    allele_order = [x for x in allele_order if alleles[x] != SPAN_DEL]
+    if len(allele_order) == 1:
+        return pd.DataFrame(
+            extract_allele_subset_from_multiallelic(multiallelic_variant, (0, allele_order[0]), record_to_nbr_dict, ref)
+        ).T
+    return pd.concat(
+        (
+            extract_allele_subset_from_multiallelic(multiallelic_variant, alleles, record_to_nbr_dict, ref)
+            for alleles in ((0, allele_order[0]), (allele_order[0], allele_order[1]))
```
doron-st, line 121: Not following this.
Are we assuming we have 3 alleles exactly? [0,1,2]
Why do we want these specific two pairs of alleles? [0,1], [1,2] or [0,2], [2,1] depending on the more confident allele

doron-st: Isn't [1,2] a multiallelic variant by itself?
In what way is it different from the original record?

ugvc/filtering/training_prep.py
```diff
@@ -0,0 +1,304 @@
...
+
+    Returns
+    -------
+    list[int]
+        List of encoded labels
+    """
+    return [encode_label(x) for x in ll]
+
+
+def encode_label(label: tuple[int, ...]) -> int:
```
doron-st, line 201: Could (0,0): 0, (0,1): 1, (1,1): 2 be more straightforward?

doron-st (approved):
OMG, I tried to review some of it...

ugvc/filtering/multiallelics.py
ilyasoifer: I am not sure. All this by-reference assignment is confusing to me.

ugvc/filtering/multiallelics.py
ilyasoifer: If there is no spanning deletion in the variant, then it is not genotyped together with the potentially spanning deletion. I am not 100% how this can happen. I added comments in this for loop that hopefully clarify

ugvc/filtering/multiallelics.py
ilyasoifer: The results need to be sorted by position. Since we first generate list of lists of spanning deletion groups and then add list of multiallelic locations, they are not sorted.

ugvc/filtering/multiallelics.py
ilyasoifer: Added description

ugvc/filtering/multiallelics.py
ilyasoifer: Yes

ugvc/filtering/multiallelics.py
ilyasoifer: Added comment.

ugvc/filtering/training_prep.py
ilyasoifer: Yes, fixed in later PR :)

ugvc/filtering/multiallelics.py
doron-st: I'm still not sure how default sorting on list of list of dataframe row (?) is going to sort by position.

PR #140, opened 2024-03-05
[BIOIN-1448] Update to the VCF filtering flow including new way of training

ilyasoifer:
OK, another kind of significant PR, following up on the previous one that affected training data prep. There are many changes that are just cosmetics (type annotations mostly), but here are three changes that I would be happy to get feedback
1. train models pipeline is now split into two: training_prep that prepares training data and train_models that receives training data and trains the model (switched to XGBoost). 
2. variant_filtering_utils changed quite a bit, specifically, removed the class for filtering models and made all the evaluations more straightforward.
3. filter_variants_pipeline was made simpler (no need to annotate the VCF in python, since the annotations from the VCF are being used)

Thanks for the help, this is the last PR that is that extensive, future ones will be more manageable


ugvc/filtering/transformers.py
```diff
@@ -0,0 +1,329 @@
...
+                ("rs_ref", default_filler, ["rs_ref"]),
+                ("rs_alt", default_filler, ["rs_alt"]),
+                ("mean_nm_ref", default_filler, ["mean_nm_ref"]),
+                ("median_nm_ref", default_filler, ["median_nm_ref"]),
+                ("mean_nm_alt", default_filler, ["mean_nm_alt"]),
+                ("median_nm_alt", default_filler, ["median_nm_alt"]),
+                ("mean_mis_ref", default_filler, ["mean_mis_ref"]),
+                ("median_mis_ref", default_filler, ["median_mis_ref"]),
+                ("mean_mis_alt", default_filler, ["mean_mis_alt"]),
+                ("median_mis_alt", default_filler, ["median_mis_alt"]),
```
doron-st, line 212: where are all of these fields come from?

doron-st (approved):
Very impressive, and hard to review.
The best I can do is test that it works I think

ugvc/pipelines/training_prep_pipeline.py
```diff
@@ -0,0 +1,120 @@
...
+        type=str,
+        required=False,
+    )
+    ap_var.add_argument(
+        "--contig_for_test",
+        help="Chromosome to split into test set, **contained** in contigs_to_read",
+        type=str,
+        required=False,
+    )
+    ap_var.add_argument("--output_prefix", help="Output HDF5 files prefix", type=str, required=True)
```
doron-st, line 68: I got many outputs in currdir, and on in output_prefix dir, might be nice to change that

ugvc/filtering/transformers.py
ilyasoifer: From my tests of filtering FFPE variants. these were added to the VCF/Dataframe by custom code. 
Currently I feel like I don't have a very good mechanism of supplying custom transformers

ugvc/pipelines/training_prep_pipeline.py
ilyasoifer: I am sorry, I could not get these results. Could you share some way to reproduce this?

doron-st: I ran with --output_prefix ../out_dir

ilyasoifer: --ouptut_prefix includes the beginning of the file name for the output (.h5) is appended to it. Should I change the name or add a parameter for the output directory?

ilyasoifer: @doron-st - sorry - I still can't reproduce this. When I run it from a different directory, I only get files under the output_prefix

doron-st:
@ilyasoifer the vcf-DF part now seems to work on my dataset.
Do you want to merge this?

ilyasoifer:
yes!

[cid:529c849f-7a9a-4a34-9b51-0b3f5e6e7965]<https://www.ultimagenomics.com/>
Ilya Soifer​​​​
Director, Bioinformatics
E:      ***@***.***
________________________________

PR #145, opened 2024-03-27
Generate CNV Figures

TammushB:


ilyasoifer (commented):
Hi, looks good, but needs type annotation and docstrings

gilhornung (changes requested):
There is no y-scale in COLO829.CNV.calls.jpeg and COLO829.dup_del.calls.jpeg

TammushB:
> There is no y-scale in COLO829.CNV.calls.jpeg and COLO829.dup_del.calls.jpeg

for COLO829.dup_del.calls.jpeg there is no actual y scale. we put y labels to differentiate between calls and GT. 
for  COLO829.CNV.calls.jpeg we actually show the copy number as the y axis and we have both labels and values on the y axis.

gilhornung (approved):

PR #153, opened 2024-04-14
Update ppmseq to 1.0.10

itamark-ug:


.github/workflows/build-ppmseq-docker-for-apl.yml
```diff
@@ -17,7 +17,7 @@ permissions:
 
 env:
   AWS_REGION: us-east-1
-  AWS_ROLE: arn:aws:iam::${{ secrets.AWS_ACCOUNT }}:role/github-actions-aws-role
```
kenissur, line 20: I don't know about this part so I trust you know what you're doing

kenissur (approved):

.github/workflows/build-ppmseq-docker-for-apl.yml
itamark-ug: There was a conception that our aws account id should be stored as a secret because this is a public repository.
However, it is not considered a secret by AWS themselves, furthermore, when using it as a secret, it does not work :-)

PR #160, opened 2024-05-17
[BIOIN-1448] Implemented filtering of VCF that changes GQ/GT/PL

ilyasoifer:
This PR updates the way we are filtering GATK VCFs. Until now we trained the filtering models using approximate ground truths. This allows marking the variant as True/False but not changing the genotype. 

In this PR I am adding the ability to use a model that is trained on the exact ground truth and in this case it is possible to not only filter the variant but also change the genotype. The model works (similar to the DeepVariant) by predicting genotype one of the three possible genotypes: hom-ref, hom-alt and het-alt. 

The main challenge that also required some code work was applying the model to multiallelic variants. The heuristic that I chose for the multiallelic variants: R,A1,A2 (the first one is the reference). 
1. Pick the "strongest" allele (e.g. A1), 
2. Combine the support for A1 and A2 into a single allele A
3. Genotype (R,A): the result is `(PL'(R,R), PL'(R,A), PL'(A,A))`
4. Genotype (A1,A2), where A1 is the reference, A2 is the "alt". 
5. Merge the support in the following way to generate `PL(R,R),PL(R,A1),PL(A1,A1),PL(R,A2),PL(A1,A2),PL(A2,A2)` tuple
```
PL(R,R) = PL'(R,R)
PL(R,A1) = PL'(R,A)
PL(A1,A1) = PL'(A,A) * PL'(A1,A1)
PL(A1,A2) = PL'(A,A) * PL(A1,A2)
PL(A2,A2) = PL(A,A) * PL(A2,A2)

Note that PL(R,A2) is zero here. 

This logic is implemented in the function `variant_filtering_utils.combine_multiallelic_spandel`

There are two main files to review, everything else is not really important (cosmetics)
`filter_variants_pipeline_main.py`
`variant_filtering_utils.py`

### CODE REVIEW

**Comment by gilhornung** [comment]
Can you please explain the context?

**Comment by ilyasoifer** [comment]
@gilhornung - updated the PR description

File: ugvc/pipelines/filter_variants_pipeline_main.py [hunk 1]
```diff
@@ -0,0 +1,201 @@
...
+                            df = df_original.copy()
+                        else:
+                            df["ml_lik"] = pd.Series([list(x) for x in scores], index=df.index)
+
+                        likelihoods = np.zeros((df.shape[0], max(df["ml_lik"].apply(len))))
+                        for i, r in enumerate(df["ml_lik"]):
+                            likelihoods[i, : len(r)] = r
+
+                        phreds = math_utils.phred(likelihoods + 1e-10)
+                        quals = np.clip(30 + phreds[:, 0] - np.min(phreds[:, 1:], axis=1), 0, None)
```
gilhornung, line 152: Interesting. I never knew this is the calculation that relates qual and PLs.

ugvc/filtering/variant_filtering_utils.py
```diff
@@ -448,3 +458,87 @@ def HMER_INDEL(df: pd.DataFrame) -> np.ndarray:
...
+        if rg.shape[0] == 1:
+            pls = split_scores[grouped_split_df.groups[g][0]]
+        else:
+            orig_alleles = original_df.at[k, "alleles"]
+            n_alleles = len(orig_alleles)
+            n_pls = n_alleles * (n_alleles + 1) / 2
+            pls = np.zeros(int(n_pls))
+            idx1 = orig_alleles.index(rg.iloc[1]["alleles"][0])
+            idx2 = orig_alleles.index(rg.iloc[1]["alleles"][1])
+            tmp = split_scores[grouped_split_df.groups[g][0]][2] * np.array(split_scores[grouped_split_df.groups[g][1]])
```
gilhornung, line 553: This and the next line are basically the "heart" of the calculation to merge the PLs.
It is kind of cryptic here with all the indices and sub-indices. Can you please document it better? What you wrote in the PR description is excellent

gilhornung: Another thing which is not clear to me - if you have more than 3 alleles, then all the "weak" alleles are stored in split_scores[grouped_split_df.groups[g][1]] ?

ilyasoifer:
@gilhornung kind reminder...

ugvc/filtering/multiallelics.py
```diff
@@ -50,7 +53,7 @@ def select_overlapping_variants(df: pd.DataFrame) -> list:
         if len(cluster) == 0:
             cluster.append(i)
         # or if we are in a deletion and the variant contains SPAN_DEL
-        elif len(cluster) > 0 and SPAN_DEL in df.iloc[i]["alleles"]:
+        elif len(cluster) > 0 and ((SPAN_DEL in df.iloc[i]["alleles"]) or (not require_star_for_spandel)):
```
gilhornung, line 56: I think you can make the code more efficient by not using a for loop. Try something like:
filter just deletions, create a new column with a range that contains all the bases in the deletion, use `df.explode` to generate a row for each base in a deletion, intersect (using merge/join) with all variants.

ugvc/filtering/variant_filtering_utils.py
```diff
@@ -94,8 +98,13 @@ def apply_model(
     tuple:
         Predictions and probabilities
     """
-    x_test_df = transformer.transform(input_df)
-    _validate_data(pd.DataFrame(x_test_df))
+    chunks = np.arange(0, input_df.shape[0], MAX_CHUNK_SIZE, dtype=int)
```
gilhornung, line 103: Why do you need the chunks?

gilhornung (approved):
Impressive work. Only cosmetic comments

ugvc/filtering/multiallelics.py
ilyasoifer: Good idea. I just think that it is anyway pretty negligible in terms of time.

ugvc/filtering/variant_filtering_utils.py
ilyasoifer: To avoid memory issues, added comment

ugvc/filtering/variant_filtering_utils.py
ilyasoifer: Right. I only try resolving the two strongest alleles. Everyone else are currenty neglected

ilyasoifer: Adding comments

PR #161, opened 2024-05-20
DATA-6490-scrna-qc-updates

inbalzelig-ugi:
Add barcode name prefix to all outputs, update histogram plots, rename entries in short table.

### DIFF
--- a/ugvc/pipelines/single_cell_qc/single_cell_qc_pipeline.py
+++ b/ugvc/pipelines/single_cell_qc/single_cell_qc_pipeline.py
@@ -1,20 +1,28 @@
-from pathlib import Path
 import subprocess
 from argparse import ArgumentParser
+from pathlib import Path
 
-
-from ugvc.pipelines.single_cell_qc.collect_statistics import collect_statistics, extract_statistics_table
+from ugvc.pipelines.single_cell_qc.collect_statistics import (
+    collect_statistics,
+    extract_statistics_table,
+)
 from ugvc.pipelines.single_cell_qc.create_plots import (
     cbc_umi_plot,
+    plot_insert_length_histogram,
     plot_mean_insert_quality_histogram,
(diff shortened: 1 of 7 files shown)


ugvc/pipelines/single_cell_qc/single_cell_qc_pipeline.py
```diff
@@ -1,37 +1,49 @@
...
-def single_cell_qc(input_files: Inputs, output_path: str, thresholds: Thresholds):
+def single_cell_qc(
+    input_files: Inputs, output_path: str, thresholds: Thresholds, barcode_name: str
+):
     """
     Run single cell qc pipeline that collects statistics, prepares parameters for report and generates report
 
     :param input_files: Inputs object with paths to input files
     :param output_path: path to output directory
     :param thresholds: Thresholds object with thresholds for qc
```
kenissur, line 31: 1. I suggest calling this "sample_name" instead of "barcode_name" since it will also run off-tool and is not necessarily a single barcode in a single run
2. Documentation for barcode_name is missing
3. We use numpy docstrings in the rest of VariantCalling
4. Do you have pre-commit configured? I think if you did then issue #2 would not have happened

ugvc/pipelines/single_cell_qc/single_cell_qc_pipeline.py
```diff
@@ -141,7 +165,7 @@ def main():
         "--star-stats", type=str, required=True, help="path to STAR stats file"
     )
     parser.add_argument(
-        "--r2-subsample", #TODO: should be renamed to insert-subsample
+        "--insert-subsample",
```
kenissur: I would just call it "insert", we are choosing to run it on a subsample

kenissur (approved):
Minor comments

PR #163, opened 2024-05-28
BIOIN-1599 scrna qc testing

inbalzelig-ugi:
* Add unit tests for scrna qc pipeline.
* use papermill python package instead of using subprocess with papermill command.


.github/workflows/python-package-conda.yml
```diff
@@ -56,5 +56,21 @@ jobs:
       - name: Test with pytest
         run: |
           source activate genomics.py3
-          python -m pytest --ignore test/unit/utils/test_db_access.py test/unit
+          python -m pytest --ignore test/unit/utils/test_db_access.py --ignore test/unit/single_cell_qc test/unit
```
ilyasoifer, line 59: Can we add 
`from __future__ import annotations`
and have the tests running in the common environment?

ilyasoifer:
@inbalzelig-ugi - could you please fix the broken build?

setup/scRNA_qc/poetry.lock
ilyasoifer, line 1: Can I ask what are all these changes and why they are needed?

ugvc/pipelines/single_cell_qc/single_cell_qc_pipeline.py
ilyasoifer, line 1: The structure of this file is not according to the conventions of our reports and pipelines. 
In general we try to separate the statistics collection (single_cell_qc_pipeline.py) and generating the report that should do as little calculations as possible. 

Also - would be nice to have the docstrings formatted in numpy style

ilyasoifer (commented):
Looks good!

.github/workflows/python-package-conda.yml
inbalzelig-ugi: I'm not sure if all the dependencies are installed in the conda env, additionally, I would like it to run with the specific poetry env to test it as well.

setup/scRNA_qc/poetry.lock
inbalzelig-ugi: We use poetry.lock file to define a specific version of a package. See here for more details: https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control

ugvc/pipelines/single_cell_qc/single_cell_qc_pipeline.py
inbalzelig-ugi: I'm not sure I follow - do you suggest moving the report generation part to a different file? 
The docstring format changed.

.github/workflows/python-package-conda.yml
ilyasoifer: I tried this locally and it seems to me that all dependencies are installed in conda env (not that surprising, it already contains all the internet :) ). 
And no problem to run the tests also in the poetry env.

setup/scRNA_qc/poetry.lock
ilyasoifer: I am a little bit concerned that we had to make this very significant update for an environment that is a month old. Is this a common occurrence or a one off?

ugvc/pipelines/single_cell_qc/single_cell_qc_pipeline.py
ilyasoifer: More precisely, the report generation is normally a different step in the WDL. So usually there is a script for collecting statistics and a script that just creates an html document that presents these (at least this is the convention that in many cases we are able to enforce)

setup/scRNA_qc/poetry.lock
inbalzelig-ugi: Don't mind this file. Poetry creates it automatically and we just need it as part of the repository in git but I didn't write a single line here.

ugvc/pipelines/single_cell_qc/single_cell_qc_pipeline.py
inbalzelig-ugi: I see, but this is not running as part of a wdl at the moment. If you want we can take this as a different task but let's focus on the current PR, unit testing.

.github/workflows/python-package-conda.yml
ilyasoifer: @inbalzelig-ugi - this comment is with you, right?

ugvc/pipelines/single_cell_qc/single_cell_qc_pipeline.py
ilyasoifer: ok, fair enough.

ilyasoifer (approved):

PR #167, opened 2024-06-20
BIOIN-1558 srsnv cv by chromosome

orihir:
## Overview
This PR adds support for training SRSNV model with a cross-validation (CV) scheme. CV is needed when SRSNV inference is used on the same featuremaps used for training, because without CV there's a danger of overfitting. 

## How to use
To use CV, need to set the new parameter `num_CV_folds` to the number of folds required. By default, this is 1, i.e., no CV (regular train-test split). Each read in the training featuremap dataframe is assigned a fold, stored in the `fold_id` column. Model `k` associated with fold `k` is trained on all data where `fold_id != k`, and validated on `fold_id == k`. By default, folds are split by chromosome, to make sure there is no leakage between training and validation sets (`fold_id`'s can be assigned randomly by setting `split_folds_by` to `random` rather than `chrom`. With random fold-splitting, at inference time an ensemble of all `num_CV_folds` models is used for inference). 

`fold_id` can have one of 3 types of values.
- A number `k` in `0, 1, ..., num_CV_folds-1`. These reads are in the training set of all models except model `k`. 
- -1. These reads are in the training set when `num_CV_folds==1`, i.e., when using regular train-test split. The test set reads in this case have `fold_id=0`. 
- `np.nan`. These reads are not included in any training set, and so (an ensemble of) all models can be used for inference. 

The new function `k_fold_predict_proba` can be used to evaluate the trained models on a dataset (it accepts a list of models, a dataset, and the name of the column containing fold_ids, and evaluates the models according to these rules. 

## Other changes in this PR
- Categorical features are defined together with all possible values, ensuring that they are consistently encoded by the model during training, validation, and inference. 
- Models are saved in a joblib file that contains relevant meta-data, like the parameters used in training (which include the mapping of chromosomes to folds, and the details of all categorical variables), and also the interpolating function that maps ML_qual to SNVQ. 
- Hyperparameters were changes to include more training rounds and use early-stopping to reduce overfitting.
- Small bug fixes, including in the calculation of the residual snv rate.


ugvc/mrd/srsnv_inference_utils.py
```diff
@@ -39,48 +45,65 @@ def _vcf_getter(variant, field):
...
-        categorical_features: list[str],
+        models,
+        categorical_features_dict: dict[str, list],
         numerical_features: list[str],
-        X_train_path: str,
+        quality_interpolation_function: Callable,
         pre_filter: str = None,
-        test_set_mrd_simulation_dataframe_file: str = None,
         low_qual_threshold: float = LOW_QUAL_THRESHOLD,
+        # new_categorical_features: list[str] = None, # <- HERE, make sure these are not used!
```
kenissur: Remove the comment please

ugvc/mrd/srsnv_inference_utils.py
```diff
@@ -164,22 +183,48 @@ def process_records(self, records: list[pysam.VariantRecord]) -> list[pysam.Vari
...
+        #     for k, model in enumerate(self.models):
+        #         if (df["fold"] == k).any():
+        #             predicted_probability[(df["fold"] == k).index, :] = model.predict_proba(
+        #                 df.loc[df["fold"] == k, features_for_model]
+        #             )  # predict on chromosomes of fold k
+        #         if df["fold"].isna().any():
+        #             predicted_probability[df["fold"].isna().index, :] += (
+        #                 model.predict_proba(df.loc[df["fold"].isna(), features_for_model]) / self.num_folds
+        #             )  # chromosomes not in any fold, predict with all models and take mean
+
```
kenissur, line 196: Please remove comments

ugvc/mrd/srsnv_inference_utils.py
```diff
@@ -164,22 +183,48 @@ def process_records(self, records: list[pysam.VariantRecord]) -> list[pysam.Vari
         )
 
         # Apply the provided model to assign a new quality value
-        predicted_probability = self.model.predict_proba(df)
+        features_for_model = self.numerical_features + self.categorical_features_names
+        if self.num_folds == 1:
+            df["fold_id"] = 0
```
kenissur: A bit petty but maybe use a constant FOLD_ID?

ugvc/mrd/srsnv_inference_utils.py
```diff
@@ -220,54 +265,53 @@ def extract_categories_from_parquet(parquet_path):
...
-    X_train_path: str,
+    model_joblib_path: str,
     params_path: str,
     model_path: str,
     test_set_mrd_simulation_dataframe_file: str,
     out_path: str,
     low_qual_threshold: float = LOW_QUAL_THRESHOLD,
     process_number: int = 0,
+    # num_folds: int = 0,
+    # chrom_folds_path: str = None,
```
kenissur: Comments

kenissur (approved):

PR #168, opened 2024-06-20
Gat/featuremap consensus

GatKrieger:
a function to pileup featuremap


ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+from collections import defaultdict
+from os.path import basename
+from os.path import join as pjoin
+
+import numpy as np
+import pysam
+
+
+def write_a_pileup_record(
+    record_dict: dict, rec_id: str, out_fh: pysam.VariantFile, header: pysam.VariantHeader, min_qual: int
```
kenissur: Please add a docstring (CoPilot makes it very fast)

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+):
+    info_fields = [
+        "X_QUAL",
+        "X_RN",
+    ]
```
kenissur: Please add these to the FeatureMapFields enum in ugvc/mrd/featuremap_utils.py an import them, then you can do something like 
`info_fields = [
        FeatureMapFields.X_QUAL.value,
        FeatureMapFields.X_RN.value,
    ]`

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    for format_field in format_fields:
+        rec.samples["SAMPLE"][format_field] = record_dict[format_field]
+
+    out_fh.write(rec)
+    return rec
+
+
+def pileup_featuremap(featuremap: str, output_dir: str, genomic_interval: str, min_qual: int):
+    """
+    Pileup featuremap vcf to a reagular vcf, save the mean quality and read count
```
kenissur: 1. Please complete the docstring, including inputs. 
2. Why is genomic_intervals a required input? According to what it says below it should be a default of "", though None is better
3. I suggest a default of min_qual=0

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    cons_dict = defaultdict(dict)
+    # process genomic interval input
+    if genomic_interval == "":
```
kenissur: Better to use None as the "no input" option

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    header.info.add(
+        "FILTERED_COUNT",
+        1,
+        "Integer",
+        "Number of reads containing this location that agree with alternative according to fitler",
+    )
+    header.info.add("X_READ_COUNT", 1, "Integer", "Number of reads containing this location")
+    header.info.add("X_QUAL", ".", "Float", "Quality of reads containing this location")
+    header.info.add("X_RN", 1, "String", "Read Name of reads containing this location, entries spearated by |")
+    header.info.add("X_AF", 1, "Float", "Allele frequency")
```
kenissur: Please use FeatureMapFields for these strings

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    # generate a new header
+    orig_header = pysam.VariantFile(featuremap).header
+    header = pysam.VariantHeader()
+    header.add_meta("fileformat", "VCFv4.2")
+    header.filters.add("FAIL", None, None, "Mean quality below threshold")
+    header.info.add(
+        "FILTERED_COUNT",
+        1,
+        "Integer",
+        "Number of reads containing this location that agree with alternative according to fitler",
```
kenissur: *filter

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    header.info.add("X_READ_COUNT", 1, "Integer", "Number of reads containing this location")
+    header.info.add("X_QUAL", ".", "Float", "Quality of reads containing this location")
+    header.info.add("X_RN", 1, "String", "Read Name of reads containing this location, entries spearated by |")
+    header.info.add("X_AF", 1, "Float", "Allele frequency")
+    header.formats.add("GT", 1, "String", "Genotype")
+    header.formats.add("DP", 1, "Integer", "Read depth")
+    header.formats.add("AD", "R", "Integer", "Allelic depths")
+    header.formats.add("VAF", 1, "Float", "Allele frequency")
+    # copy contigs from original header
+    for contig in orig_header.contigs:
```
kenissur, line 200: Are you sure this is the only thing needed from orig_header? I am not saying otherwise, just to be sure. For example, maybe it would be good to save a list of the individual aggregated features instead of just the quality?

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+        header.contigs.add(contig)
+    header.add_sample("SAMPLE")
+
+    # open an output file
+    output_vcf = pjoin(
+        output_dir, basename(featuremap).replace(".vcf.gz", f".{genomic_interval.replace(':', '-')}.pileup.vcf.gz")
```
kenissur: 1. If genomic_interval=="" you will get an output with "<>..pileup.vcf.gz"
2. I suggest taking the filename as input and not the path, since you are generating one output file

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    out_fh = pysam.VariantFile(output_vcf, "w", header=header)
+
+    cons_dict = defaultdict(dict)
+    with pysam.VariantFile(featuremap) as f:
+        for rec in f.fetch(chrom, start, end):
+            rec_id = "-".join([str(x) for x in (rec.chrom, rec.pos, rec.ref, rec.alts[0])])
+            if "FILTERED_COUNT" not in cons_dict[rec_id]:
+                if len(cons_dict.keys()) > 1:
+                    # write to file
+                    prev_key = list(cons_dict.keys())[0]
```
kenissur: I'm concerned about this line, because I don't think the order of keys is guaranteed. I think you need an OrderedDict for that. 

Also, is it possible you are counting the last entry twice? You require "len(cons_dict.keys()) > 1" before you assign cons_dict[rec_id], so I think you will do another iteration of the lines that add to cons_dict[rec_id] before entering the if-condition that pops the prev_key.

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    # open an output file
+    output_vcf = pjoin(
+        output_dir, basename(featuremap).replace(".vcf.gz", f".{genomic_interval.replace(':', '-')}.pileup.vcf.gz")
+    )
+    out_fh = pysam.VariantFile(output_vcf, "w", header=header)
+
+    cons_dict = defaultdict(dict)
+    with pysam.VariantFile(featuremap) as f:
+        for rec in f.fetch(chrom, start, end):
+            rec_id = "-".join([str(x) for x in (rec.chrom, rec.pos, rec.ref, rec.alts[0])])
```
kenissur: Why not use "(rec.chrom, rec.pos, rec.ref, rec.alts[0])" instead of making it into a string? You can still compare it and use it as a dictionary key

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+    format_fields = ["GT", "AD", "DP", "VAF"]
+
+    rec = header.new_record()
+    rec.chrom = rec_id.split("-")[0]
+    rec.pos = int(rec_id.split("-")[1])
+    rec.ref = rec_id.split("-")[2]
+    rec.alts = [rec_id.split("-")[3]]
+    rec.id = "."
+    rec.qual = np.mean(record_dict["X_QUAL"])
```
kenissur: I don't think the mean is the best aggregation, since additional supporting reads can decrease it. I think even the highest qual would be better, until we have a more refined algorithm.

Maybe it's also good to have some support for a quality aggregation function as an input.

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    record_dict["X_RN"] = "|".join(record_dict["X_RN"])
+    curr_filter = "PASS" if rec.qual >= min_qual else "FAIL"
```
kenissur: What about using PASS/SingleRead/LowQual as the filter, with SingleRead being just one read over some quality?

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    record_dict["VAF"] = record_dict["FILTERED_COUNT"] / record_dict["DP"]
+    rec.filter.add(curr_filter)
+
+    for info_field in info_fields:
+        rec.info[info_field] = record_dict[info_field]
+
+    record_dict["GT"] = (0, 1)
+    record_dict["AD"] = [record_dict["FILTERED_COUNT"], record_dict["DP"] - record_dict["FILTERED_COUNT"]]
```
kenissur: I think the first entry is the reference and the second is the alt

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    for format_field in format_fields:
+        rec.samples["SAMPLE"][format_field] = record_dict[format_field]
```
kenissur: I suggest adding sample_name="SAMPLE" as an input and using it instead of a heardcoded string

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+def write_a_pileup_record(
+    record_dict: dict, rec_id: str, out_fh: pysam.VariantFile, header: pysam.VariantHeader, min_qual: int
+):
+    info_fields = [
+        "X_QUAL",
+        "X_RN",
+    ]
+
+    format_fields = ["GT", "AD", "DP", "VAF"]
```
kenissur: Maybe add to ugvc/dna/format.py? Not critical

ugvc/mrd/featuremap_consensus_utils.py
GatKrieger: But we are missing "X_QUAL" in FeatureMapFields(Enum)

kenissur:
It didn’t exist before, I suggest you add it there

***@***.***<https://www.ultimagenomics.com/>
Itai Rusinek​​​​
Staff Researcher
E:
***@***.***

ugvc/mrd/featuremap_consensus_utils.py
GatKrieger: not sure what you meant

kenissur:
The string above says "fitler", just a typo

ugvc/mrd/featuremap_consensus_utils.py
GatKrieger: I think I solved it now

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+        out_fh (pysam.VariantFile): A pysam file handle of the output vcf file
+        header (pysam.VariantHeader): A pysam VarianyHeader pbject of the output vcf file
+        min_qual (int): The minimum quality threshold
+        qual_agg_func (function): The function to aggregate the quality scores (default: np.max)
+        sample_name (str): The name of the sample (default: "SAMPLE")
+    Output:
+        rec (pysam.VariantRecord): A pysam VariantRecord object
+
+    """
+    format_fields = [GT, AD, DP, VAF]
```
kenissur, line 78: Why not make it a keyword argument as well?

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+    record_dict[GT] = (0, 1)
+    record_dict[AD] = [
+        record_dict[DP] - record_dict[FeatureMapFields.FILTERED_COUNT.value],
+        record_dict[FeatureMapFields.FILTERED_COUNT.value],
+    ]
+    for format_field in format_fields:
+        rec.samples[sample_name][format_field] = record_dict[format_field]
+    # FILTER field
+    if rec.qual < min_qual:
+        curr_filter = "LowQual"
```
kenissur: Please make PASS, LowQual and SingleRead constants.

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+    # write to file
+    out_fh.write(rec)
+    return rec
+
+
+def pileup_featuremap(
+    featuremap: str, output_vcf: str, genomic_interval: str = None, min_qual: int = 0, sample_name: str = "SAMPLE"
+):
+    """
+    Pileup featuremap vcf to a reagular vcf, save the mean quality and read count
```
kenissur: reagular -> regular
mean quality -> aggregated quality

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+    Output:
+        output_vcf (str): The output vcf file
+    """
+    cons_dict = defaultdict(dict)
+    # process genomic interval input
+    if genomic_interval is None:
+        chrom = None
+        start = None
+        end = None
+    if genomic_interval is not None:
```
kenissur: "if genomic_interval is not None:" -> "else"

Please add this check:
assert re.match(r'^\w+:\d+-\d+$', genomic_interval), "Input genomic_interval format should be 'chrom:start-end'"

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+    # generate a new header
+    orig_header = pysam.VariantFile(featuremap).header
+    header = pysam.VariantHeader()
+    header.add_meta("fileformat", "VCFv4.2")
+    # header.filters.add("PASS", None, None, "Aggregated quality above threshold, more than a single alternative read")
+    header.filters.add(
+        "SingleRead", None, None, "Aggregated quality above threshold, Only a single read agrees with alternative"
+    )
+    header.filters.add("LowQual", None, None, "Aggregated quality below threshold")
+    for field in fields_to_collect["numeric_aggregated_fields"]:
```
kenissur: if "field" is not in "fields_to_collect", do you want it to continue without it with just a warning, or raise an error? In any case please check that "field in orig_header.info" and if False handle accordingly

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+        header.info.add(
+            field, orig_header.info[field].number, orig_header.info[field].type, orig_header.info[field].description
+        )
+    for field in fields_to_collect["numeric_array_fields"]:
+        header.info.add(field, ".", orig_header.info[field].type, orig_header.info[field].description)
+    for field in fields_to_collect["string_list_fields"]:
+        header.info.add(field, "1", orig_header.info[field].type, orig_header.info[field].description)
+    # exceptions: X_QUAL
+    header.info.add(FeatureMapFields.X_QUAL.value, ".", "Float", "Quality of reads containing this location")
```
kenissur, line 194: Maybe "Aggregated quality" in the docstring

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+    for field in fields_to_collect["numeric_array_fields"]:
+        if field in record_dict:
+            rec.info[field] = list(record_dict[field])
+    for field in fields_to_collect["string_list_fields"]:
+        if field in record_dict:
+            rec.info[field] = "|".join(record_dict[field])
+
+    # FORMAT fields to aggregate
+    record_dict[VAF] = record_dict[FeatureMapFields.FILTERED_COUNT.value] / record_dict[DP]
+    record_dict[GT] = (0, 1)
```
kenissur, line 116: Why set the genotype? I guess if there's no other choice a HET variant makes the most sense but isn't it better not to set it at all?

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+                cons_dict[rec_id][field] += 1
+            # exceptions: X_QUAL
+            cons_dict[rec_id]["X_QUAL"] = np.append(cons_dict[rec_id]["X_QUAL"], rec.qual)
+            for field in fields_to_collect["numeric_array_fields"]:
+                cons_dict[rec_id][field] = np.append(cons_dict[rec_id][field], rec.info[field])
+            for field in fields_to_collect["string_list_fields"]:
+                cons_dict[rec_id][field] += [rec.info[field]]
+            for field in fields_to_collect["fields_to_write_once"]:
+                cons_dict[rec_id][field] = rec.info[field]
+            cons_dict[rec_id][DP] = rec.info[FeatureMapFields.READ_COUNT.value]
```
kenissur: Should now be updated to FILTERED_COUNT, or better yet "sum(cons_dict[rec_id][AD])"

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+from collections import defaultdict
+
+import numpy as np
+import pysam
+
+from ugvc.dna.format import AD, DP, GT, VAF
+from ugvc.mrd.featuremap_utils import FeatureMapFields
+
+fields_to_collect = {
+    "numeric_aggregated_fields": [FeatureMapFields.FILTERED_COUNT.value],
```
kenissur: I now realize that this is the same as X_FILTERED_COUNT, that has a different meaning in a regular FeatureMap, which is confusing. Could you change this to something like "ALT_COUNT"?

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+                    cons_dict[rec_id][field] = 0
+                # exceptions: X_QUAL
+                cons_dict[rec_id]["X_QUAL"] = np.array([])
+                for field in fields_to_collect["numeric_array_fields"]:
+                    cons_dict[rec_id][field] = np.array([])
+                for field in fields_to_collect["string_list_fields"]:
+                    cons_dict[rec_id][field] = []
+            # update the record
+            for field in fields_to_collect["numeric_aggregated_fields"]:
+                cons_dict[rec_id][field] += 1
```
kenissur: Are "numeric_aggregated_fields" by definition counters or records? It looks like it to me, so in this case it should just be one variable rather than a list.

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+            # exceptions: X_QUAL
+            cons_dict[rec_id]["X_QUAL"] = np.append(cons_dict[rec_id]["X_QUAL"], rec.qual)
```
kenissur: Please use the constant for X_QUAL

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+            for field in fields_to_collect["numeric_array_fields"]:
+                cons_dict[rec_id][field] = np.append(cons_dict[rec_id][field], rec.info[field])
+            for field in fields_to_collect["string_list_fields"]:
+                cons_dict[rec_id][field] += [rec.info[field]]
+            for field in fields_to_collect["fields_to_write_once"]:
```
kenissur: Do you want to make sure they are all identical? If they are not you will get the last one which seems arbitrary

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+                cons_dict[rec_id][field] = rec.info[field]
+            cons_dict[rec_id][DP] = rec.info[FeatureMapFields.READ_COUNT.value]
+            prev_key = rec_id
```
kenissur: It probably does the same but I would do this right after "cons_dict.pop(prev_key)"

kenissur (commented):
It looks excellent! Very well structured. I had some hopefully last comments, most of them very small, and I think we can finish.

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+        FeatureMapFields.X_EDIST.value,
+        FeatureMapFields.X_LENGTH.value,
+        FeatureMapFields.X_MAPQ.value,
+        FeatureMapFields.X_INDEX.value,
+        FeatureMapFields.X_FC1.value,
+        FeatureMapFields.X_FC2.value,
+        FeatureMapFields.MAX_SOFTCLIP_LENGTH.value,
+        FeatureMapFields.X_FLAGS.value,
+        # "a3",
+        # "rq"
```
kenissur: Please remove comments

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+    ],
+    "string_list_fields": [
+        FeatureMapFields.X_RN.value,
+        FeatureMapFields.X_CIGAR.value,
+        "rq"
+        # FeatureMapFields.IS_FORWARD.value,
+        # FeatureMapFields.IS_DUPLICATE.value,
+        # "tm"
```
kenissur: Please remove comments

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,211 @@
...
+        FeatureMapFields.NEXT_1.value,
+        FeatureMapFields.NEXT_2.value,
+        FeatureMapFields.NEXT_3.value,
+    ],
+}
+
+
+def write_a_pileup_record(
+    record_dict: dict,
+    rec_id: str,
```
kenissur: rec_id is a tuple, isn't it?

ugvc/mrd/featuremap_consensus_utils.py
GatKrieger: I looked for it, there's no docstring for QUAL in vcfs.

ugvc/mrd/featuremap_consensus_utils.py
GatKrieger: it's a counter. Is there something I should change?

ugvc/mrd/featuremap_consensus_utils.py
GatKrieger: I updated it to FILTERED_COUNT

ugvc/mrd/featuremap_consensus_utils.py
kenissur: I just meant that there is no reason to have a list of variables if they would all be counters, it can be just one variable

ugvc/mrd/featuremap_consensus_utils.py
GatKrieger: I changed it to simply "rec_counter", in the end it's written into "AD[1]"

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -21,36 +21,40 @@
...
         FeatureMapFields.NEXT_2.value,
         FeatureMapFields.NEXT_3.value,
     ],
+    "boolean_fields_to_write_once": [
+        FeatureMapFields.IS_CYCLE_SKIP.value,
+    ],
 }
 
 
+class FeatureMapFilters(Enum):
```
kenissur: That's good, but could you put it in ugvc/mrd/featuremap_utils.py and import it? ugvc/mrd/srsnv_inference_utils.py has similar definitions so it should import the same structure rather than have this duplicated

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -90,23 +93,36 @@ def write_a_pileup_record(
     for field in fields_to_collect["string_list_fields"]:
         if field in record_dict:
             rec.info[field] = "|".join(record_dict[field])
+    for field in fields_to_collect["boolean_fields"]:
+        if field in record_dict:
+            rec.info[field] = "|".join(["T" if f else "F" for f in record_dict[field]])
+    for field in fields_to_collect["fields_to_write_once"]:
+        if field in record_dict:
+            if [f == record_dict[field][0] for f in record_dict[field]].count(False) == 0:
```
kenissur: 1. This syntax is a bit confusing, I think it's better to use "any" or "all", e.g "if all([f == record_dict[field][0] for f in record_dict[field]])"
2. I don't understand the logic here, could you explain? You are checking that the list is just one unique value? If it's not, what happens?

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -90,23 +93,36 @@ def write_a_pileup_record(
...
+                rec.info[field] = record_dict[field][0]
+    for field in fields_to_collect["boolean_fields_to_write_once"]:
+        if field in record_dict:
+            if [f == record_dict[field][0] for f in record_dict[field]].count(False) == 0:
```
kenissur: Same comment as above

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -90,23 +93,36 @@ def write_a_pileup_record(
...
+                rec.info[field] = record_dict[field][0]
 
     # FORMAT fields to aggregate
-    record_dict[VAF] = record_dict[FeatureMapFields.FILTERED_COUNT.value] / record_dict[DP]
-    record_dict[GT] = (0, 1)
+    record_dict[DP] = rec.info[FeatureMapFields.FILTERED_COUNT.value]
+    record_dict[VAF] = record_dict["rec_counter"] / record_dict[DP]
+    # record_dict[GT] = (0, 1)
```
kenissur: Please remove the comment

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -155,6 +166,7 @@ def pileup_featuremap(
         chrom = genomic_interval_list[0]
         start = int(genomic_interval_list[1].split("-")[0])
         end = int(genomic_interval_list[1].split("-")[1])
+    # print(chrom, start, end)
```
kenissur: remove comment please

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -237,8 +251,64 @@ def pileup_featuremap(
...
+                    + str(start)
+                    + "_"
+                    + str(end)
+                    + ".int_list.vcf.gz",
+                )
+                pileup_featuremap(featuremap, curr_output_vcf, genomic_interval, min_qual, sample_name, qual_agg_func)
+        # merge the output vcfs
+        vcfs = [pjoin(temp_dir, f) for f in os.listdir(temp_dir) if f.endswith(".vcf.gz")]
+        vcf_str = " ".join(vcfs)
+        cmd = f"bcftools concat {vcf_str} | bcftools sort - -Oz -o {output_vcf} && bcftools index {output_vcf}"
```
kenissur: 1. bcftools index **-t**
2. Do you need to sort? I think an interval_list file is required to be sorted already - "The intervals MUST be sorted by coordinate (in increasing order) within contigs; and the contigs must be sorted in the same order as in the sequence dictionary." https://gatk.broadinstitute.org/hc/en-us/articles/360035531852-Intervals-and-interval-lists

ugvc/mrd/featuremap_consensus_utils.py
GatKrieger: 1. OK
2. if not, it doesn't write the entry
do you suggest otherwise? to raise an error?

kenissur: I think raising an error in this case is better, since you do expect them all to be the same

ugvc/mrd/featuremap_consensus_utils.py
```diff
@@ -0,0 +1,323 @@
...
+    # copy contigs from original header
+    for contig in orig_header.contigs:
+        header.contigs.add(contig)
+    header.add_sample(sample_name)
+
+    # open an output file
+    out_fh = pysam.VariantFile(output_vcf, "w", header=header)
+
+    # sort the vcf file prior to reading
+    sorted_featuremap = featuremap.replace(".vcf.gz", ".sorted.vcf.gz")
```
kenissur: This is not a safe mechanism, you could be overriding something. Specifically if you get a .vcf file it will try to overwrite itself. Using tempfile is generally best IMO, e.g.
```
import tempfile

with tempfile.NamedTemporaryFile(mode='w+t') as sorted_featuremap:
    # do stuff

```
It's also deleted automatically in the end.

kenissur (approved):
Looks great, please just address the one comment and you can merge. Please rebase onto master instead of merging, unless there's a specific reason not to.

ugvc/mrd/featuremap_consensus_utils.py
GatKrieger: it doesn't work this way, so I used a temp directory.

PR #170, opened 2024-06-21
Added support to the genomics.py3 env in single_cell_qc

ilyasoifer:
1. Added from __future__ import annotations to support python 3.8 that is installed in genomics.py3
2. Reformatted the code using pre-commit standard of the repo


ilyasoifer:
Made redundant by #187

PR #171, opened 2024-06-26
Update scrna qc build

inbalzelig-ugi:


setup/scRNA_qc/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "scrna_qc"
-version = "1.0.12"
+version = "1.0.13"
```
noam-ultima, line 3: if you're just changing the build, I think it isn't mandatory to change the version.
if it's related to a wdl, don't forget to update globals in terra_pipeline repo

.github/workflows/build-scrna_qc-docker-for-apl.yml
```diff
@@ -30,12 +32,16 @@ jobs:
...
           role-to-assume: ${{ env.AWS_ROLE }}
           aws-region: ${{ env.AWS_REGION }}
-          mask-aws-account-id: false
 
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: true
 
+      - uses: benjlevesque/short-sha@v3.0
```
noam-ultima: look at latest version of olink build
https://github.com/Ultimagen/data-analytics-pipeline/blob/DATA-6612-create-wdl-for-trimmer-and-olink/.github/workflows/build_olink.yaml
I removed this section
And also changed below how we use `github.sha` instead of `steps.short-sha.outputs.sha`

noam-ultima (changes requested):

setup/scRNA_qc/pyproject.toml
inbalzelig-ugi: you are right but I don't want to change it back to avoid issues in next versions

noam-ultima (approved):

itamark-ug (approved):

PR #181, opened 2024-07-10
[bioin-1723] Improve robustness of signature calculation

ilyasoifer:
This should improve the robustness (install the genomes once rather than installing them at each signature calculation)


gilhornung (approved):
Thank you for taking care of these bugs

PR #203, opened 2024-09-02
Gat/update gatk

GatKrieger:
The terra_pipeline branch: gat/update_gatk
tests: https://github.com/Ultimagen/terra_pipeline/actions/runs/10652645799

### DIFF
--- a/Dockerfile.jbvc
+++ b/Dockerfile.jbvc
@@ -33,6 +33,14 @@ RUN source ~/.bashrc && \
     cd /VariantCalling && \
     pip install .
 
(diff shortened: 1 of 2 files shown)


ilyasoifer:
@GatKrieger - sorry, I am concerned about removing gatk from genomics.py3 environment. 
This is too big of a change: everyone who builds genomics.py3 would have to install GATK separately. 

We can either wait until GATK 4.6 is released into conda (1-2 months) or see if you don't need to use PrintReads within genomics.py3

PR #209, opened 2024-09-11
BIOIN-1668-srsnv-optimizations

inbalzelig-ugi:
1. use bcftools to merge vcf (instead of pysam).
2. minor pandas updates to speed up operations.
3. Run annotate_featuremap with multiple threads


ugbio_utils
avigailmo, line 1: any reason for this update of ugbio submodule?

kenissur (approved):

PR #214, opened 2024-10-02
[BIOIN-1832] Helper script that collects hompolymer locations in reference genome

ilyasoifer:
Added a script that collects a sample of locations of homopolymers in a reference genome. 
To be used for collecting error statistics in the future


ilyasoifer:
@gilhornung , @doron-st - review much appreciated.

gilhornung (approved):
Looks good. But why do you need a random sample?

PR #215, opened 2024-10-06
methylation qc docker to gcp

giuliovn:


GatKrieger (approved):

avigailmo:
@GatKrieger I think next step is to move the methylation qc to ugbio-utils.
Now, after @inbalzelig-ugi last change I think it's a good time to do this

PR #216, opened 2024-10-07
[BIOIN-1808] delete mrd modules test resources

avigailmo:


kenissur (commented):
Did all these files pass to ugbio?

avigailmo:
> Did all these files pass to ugbio?

sure!
I just didn't want to delete them in the same pr - that was huge also without them

kenissur (approved):

PR #257, opened 2026-01-18
BIOIN-2609 No ground truth report was failing when vaf is a scalar and not tuple

ilyasoifer:
Fixed minor non-robust points in noGT report that failed when VAF was an integer rather than a tuple


gilhornung (approved):
I wonder why i didn't get this bug

